Issue #8, opened 2019-10-16
Decrease verbosity


fredrikekre:
In the web interface you can leave a comment with the approve, does the API allow that? Would send a single notification instead of both https://github.com/JuliaRegistries/General/pull/4422#issuecomment-542615772 and the approve.

fredrikekre:
Also; maybe this comment is superfluous: https://github.com/JuliaRegistries/General/pull/4422#issuecomment-542628805 ?

DilumAluthge:
> Also; maybe this comment is superfluous: [JuliaRegistries/General#4422 (comment)](https://github.com/JuliaRegistries/General/pull/4422#issuecomment-542628805) ?

I agree. Actually I think that I put that comment in as a debugging purpose, and never ended up taking it out. I’ll take that comment out.

DilumAluthge:
> In the web interface you can leave a comment with the approve, does the API allow that? Would send a single notification instead of both [JuliaRegistries/General#4422 (comment)](https://github.com/JuliaRegistries/General/pull/4422#issuecomment-542615772) and the approve.

This should be doable. Currently I do actually use the comment body of the approval to store the SHA that was approved. You can’t see it, because I hide it in an HTML comment. So it should be easy to just use the comment body of the approval for everything. I’ll just need to make sure that it doesn’t break anything.

fredrikekre:
> This should be doable. Currently I do actually use the comment body of the approval to store the SHA that was approved. You can’t see it, because I hide it in an HTML comment.

Turns out that on the mobile page you can see the comment but it is displayed as empty (empty white block) whereas on GH it is not shown at all.

DilumAluthge:
Huh, interesting.

At any rate, I think the best approach will be to put the text of the “congratulations you met the guidelines” message inside the approval body. And then I’ll still have the hidden (commented out) message at the end, but it won’t be visible.

PR #12, opened 2019-10-16
Remove superfluous comment

DilumAluthge:
This solves part (but not all) of #8.

Specifically, it removes a superfluous comment, as discussed here: https://github.com/JuliaRegistries/GeneralRegistryCI.jl/issues/8#issuecomment-542634516

### DIFF
--- a/src/AutoMerge/cron.jl
+++ b/src/AutoMerge/cron.jl
@@ -179,7 +179,7 @@ function travis_cron_or_api_build(pr::GitHub.PullRequest,
                                 @info(string("Pull request: $(pr_number). ",
                                              "Type: $(pr_type). ",
                                              "Decision: merge now."))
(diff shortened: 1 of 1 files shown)


fredrikekre:
https://github.com/JuliaRegistries/General/pull/4434#issuecomment-542939547 ?

DilumAluthge:
That is bizarre.

DilumAluthge:
This makes zero sense.

DilumAluthge:
Ah no it makes perfect sense

DilumAluthge:
Because I made a mistake

DilumAluthge:
We use different logic for new package vs new version. And I only fixed new package. Didn’t fix new version.

DilumAluthge:
Will be fixed by https://github.com/JuliaRegistries/RegistryCI.jl/pull/17

PR #77, opened 2019-10-29
Put approve-comment in the review

### LINKED ISSUE #8: Decrease verbosity

fredrikekre:
Put approve-comment in the review instead of as a separate comment, fixes #8.

### DIFF
--- a/src/AutoMerge/github.jl
+++ b/src/AutoMerge/github.jl
@@ -1,16 +1,20 @@
 function approve!(repo::GitHub.Repo,
                   pr::GitHub.PullRequest,
                   pr_head_commit_sha_to_approve::String;
-                  auth::GitHub.Authorization)
+                  auth::GitHub.Authorization,
+                  body::String="")
     whoami = username(auth)
     if pr.user.login == whoami
     else
         repo_full_name = full_name(repo)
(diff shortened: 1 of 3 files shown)


DilumAluthge:
I think this is fine for the short-term. It allows us to reduce the number of notifications (on a PR that passes automerge) from 2 to 1.

My goal is to eliminate the PR review approval altogether. So when I do that, I'll remove the `approve` and add back the comment. That's why I commented out those lines, instead of deleting them. The number of notifications will remain 1, it'll just be 1 comment and 0 reviews instead of 0 comments and 1 review.

fredrikekre:
Alright, I thought we should still approve even though what we look for is that status.

DilumAluthge:
Oh I see what you mean.

Yes we can still leave the approval if you find that helpful!

DilumAluthge:
In that case, I should actually remove those commented-out lines, to make the code cleaner/more readable.

fredrikekre:
Can do that later.

PR #79, opened 2019-10-29
Hide `[noblock]` footer in a comment

christopher-dG:
As mentioned on Slack; feel free to discuss further if it's not unanimously wanted.
My vote is hide it, since there are already ample instructions about the blocking mechanism in the comment.


KristofferC:
Seems like it is also incorporated in https://github.com/JuliaRegistries/RegistryCI.jl/pull/78.

christopher-dG:
Oops! Didn't read that diff closely enough, I thought it was just in the actual instruction content.

PR #81, opened 2019-10-29
Add the `always_assert` function

DilumAluthge:
The documentation for the `@assert` macro says:
> Warning: An assert might be disabled at various optimization levels.

This pull request adds the `always_assert` function. `always_assert` is like `@assert`, except that `always_assert` will always run and will never be disabled.


DilumAluthge:
I like the idea. What would it look like?

fredrikekre:
```julia
function assert(cond)
    cond || throw(Base.AssertionError())
    return nothing
end
```

PR #89, opened 2019-10-31
Add a bit of guidance to the auto-merge comment

mschauer:


src/AutoMerge/util.jl
```diff
@@ -26,6 +26,16 @@ function _clone_repo_into_dir(url::AbstractString, repo_dir)
     return repo_dir
 end
 
+function _comment_guide()
+    result = string("\n\n",
+                    "Note: If your package works for the current version `x.y.z` of a dependency `foo`, ",
+                    "then a compat entry `foo = x.y.z` implies a compatibility upper bound ",
+                    "for packages following semver. You can additionally include earlier versions",
+                    "your package is compatible with. ",
+                    "See [https://julialang.github.io/Pkg.jl/v1/compatibility/.] for details.")
```
StefanKarpinski: A multiline `"""` string seems like it would be much easier to read/write here.

mschauer: This follows the style of the surrounding code it will look a bit odd if we change it here.

fredrikekre: I believe the reason is that we don't want newlines.

StefanKarpinski: Ah, ok. I guess that makes sense.

PR #90, opened 2019-11-01
Explicitly use caret notation for all compat entries

DilumAluthge:


KristofferC:
Ok but why :p.

DilumAluthge:
Fair enough :P 

Mainly to be consistent with the other compat bounds

Although I guess we could go the other way and get rid of all carets

PR #91, opened 2019-11-01
Be more concise/readable in compat bounds

DilumAluthge:


DilumAluthge:
@kristofferc What do you think? It actually does look a bit more readable this way

PR #96, opened 2019-11-06
Do not consider `PkgA = "0"` to be an upper-bounded compat range

DilumAluthge:
Before this PR:
- `Pkg.Types.VersionBound("*")` is not an upper bound
- Every other `VersionBound` is an upper bound

After this PR:
- `Pkg.Types.VersionBound("*")` is not an upper bound
- `Pkg.Types.VersionBound("0")` is not an upper bound
- Every other `VersionBound` is an upper bound

### DIFF
--- a/src/AutoMerge/semver.jl
+++ b/src/AutoMerge/semver.jl
@@ -45,7 +45,7 @@ function julia_compat(pkg::String, version::VersionNumber, registry_path::String
 end
 
 function _has_upper_bound(r::Pkg.Types.VersionRange)
-    return r.upper != Pkg.Types.VersionBound("*")
+    return r.upper != Pkg.Types.VersionBound("*") && r.upper != Pkg.Types.VersionBound("0")
 end
 
 function range_did_not_narrow(r1::Pkg.Types.VersionRange, r2::Pkg.Types.VersionRange)


tkf:
IIUC, this rule still can be avoided by saying `PkgA = "9999"`?  If you want to be _super_ strict about this, maybe just disallow non-existing major version for post-v1 releases and non-existing minor version for pre-v1?

DilumAluthge:
> IIUC, this rule still can be avoided by saying `PkgA = "9999"`? If you want to be _super_ strict about this, maybe just disallow non-existing major version for post-v1 releases and non-existing minor version for pre-v1?

No. The compat entry only includes versions of the form `9999.y.z`.

```julia
julia> import Pkg

julia> a = "9999"
"9999"

julia> b = Pkg.Types.semver_spec(a)
VersionSpec("9999")

julia> v"0.1" in b
false

julia> v"0.2" in b
false

julia> v"1.1" in b
false

julia> v"1.2" in b
false

julia> v"9999.1" in b
true

julia> v"9999.2" in b
true

julia> v"10000.1" in b
false

julia> v"10000.2" in b
false
```

tkf:
Ah, of course. I forgot about that. Initially, I was thinking `PkgA = "0.1, 0.2, ..., 0.999"` but then tried to simply it (and failed).

Anyway, does the rule in this PR reject `PkgA = "0.1, 0.2, ..., 0.999"`? (the `...` means all the intermediate versions are written in there)

DilumAluthge:
> Anyway, does the rule in this PR reject `PkgA = "0.1, 0.2, ..., 0.999"`? (the `...` means all the intermediate versions are written in there)

No, we allow that. If you write out all of those breaking versions by hand and claim to us that you are compatible with all of those breaking versions, we will believe that you are telling the truth.

tkf:
What about `PkgA = "< 1.0.1"`?

DilumAluthge:
> What about `PkgA = "< 1.0.1"`?

Yeah we accept that as having an upper-bound. I don't really know how we'd fix that though.

```julia
julia> a = Pkg.Types.semver_spec("< 1.0.1")
VersionSpec("0.0.0-1.0.0")

julia> _has_upper_bound(a.ranges[1])
true

julia> v"0.1" in a
true

julia> v"0.2" in a
true

julia> v"0.3" in a
true

julia> v"0.4" in a
true

julia> v"0.5" in a
true

julia> v"1.0" in a
true

julia> v"1.1" in a
false
```

tkf:
How about looking at the registered versions of PkgA and reject any bounds that include non-existing major versions?

DilumAluthge:
> How about looking at the registered versions of PkgA and reject any bounds that include non-existing major versions?

But then `PkgA = "< 0.9999999"` would still work.

tkf:
Right, so that's why in my original proposal I suggest to treat pre-v1 differently https://github.com/JuliaRegistries/RegistryCI.jl/pull/96#issuecomment-550594433 (I was sloppy in the previous comment)

DilumAluthge:
It would be easiest to just disallow using `<` `>`, `<=`, `>=` in compat entries.

00vareladavid:
I agree that a compat entry such as `"< 9000"` is essentially meaningless and should not be considered valid

DilumAluthge:
Bump @fredrikekre. What are your thoughts on this?

DilumAluthge:
It doesn't seem like this is the best solution, as there are many flaws. I'm going to close this. Maybe we can come up with a better solution.

PR #104, opened 2019-11-10
Better warning when a patch changes Julia compat

asinghvi17:


DilumAluthge:
Don’t use a multiline string. It inserts a bunch of newlines that we don’t want.

Use something like `string("foo ", "bar ", "baz.")`

PR #115, opened 2019-11-23
Fix bug that happens if package doesn’t start with uppercase letter

DilumAluthge:


DilumAluthge:
Yeah good call.

https://github.com/JuliaRegistries/RegistryCI.jl/issues/116

DilumAluthge:
> This should use the path in `Registry.toml`.

This will be implemented in https://github.com/JuliaRegistries/RegistryCI.jl/pull/410

PR #121, opened 2019-12-04
Run all package-loads under xvfb

timholy:
May fix Gtk and dependent packages (https://github.com/JuliaRegistries/General/pull/6007, https://github.com/JuliaRegistries/General/pull/5890). Hopefully it isn't much of a burden for other packages.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>"]
-version = "0.3.1"
+version = "0.4.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/AutoMerge/guidelines.jl
+++ b/src/AutoMerge/guidelines.jl
@@ -174,6 +174,10 @@ function meets_version_can_be_loaded(working_directory::String,
               env = Dict("PATH" => ENV["PATH"],
                          "PYTHON" => "",
                          "JULIA_DEPOT_PATH" => tmp_dir))
+    # GUI toolkits may need a display just to load the package
+    if (xvfb = Sys.which("xvfb-run")) !== nothing
+        pushfirst!(cmd.exec, xvfb)
+    end
     @info("Attempting to install the package")
     cmd_ran_successfully = success(pipeline(cmd, stdout=stdout, stderr=stderr))
     rm(tmp_dir; force = true, recursive = true)


.travis.yml
```diff
@@ -30,6 +30,13 @@ notifications:
 os:
   - linux
 
+addons:
+    apt:
+        packages:
+            - xvfb
+            - xauth
+            - libgtk-3-0
```
DilumAluthge: It looks like you are trying to add these dependencies to Travis?

We actually don’t run the auto merge on Travis; we run it on GH Actions.

So you would need to install these dependencies in the GH Actions workflow file here: https://github.com/JuliaRegistries/General/blob/master/.github/workflows/automerge.yml

fredrikekre:
Perhaps make it conditional on if `xvfb` is installed, i.e. something like
```julia
cmd = Cmd(`$(Base.julia_cmd()) -e $(code)`;
          env = Dict("PATH" => ENV["PATH"],
          "PYTHON" => "",
          "JULIA_DEPOT_PATH" => tmp_dir))

if (xvfb = Sys.which("xvfb-run")) !== nothing
    pushfirst!(cmd.exec, xvfb)
end
```
and then we make sure to install it in GH Actions as @DilumAluthge said?

timholy:
Thanks for the suggestions! Both have been implemented.

timholy:
The previous test (https://travis-ci.com/JuliaRegistries/RegistryCI.jl/jobs/262909814) failed in a `key "BCBI_TEST_USER_GITHUB_TOKEN" not found`, I assume this is nothing I need to worry about?

fredrikekre:
> The previous test (https://travis-ci.com/JuliaRegistries/RegistryCI.jl/jobs/262909814) failed in a key "BCBI_TEST_USER_GITHUB_TOKEN" not found, I assume this is nothing I need to worry about?

Likely because secret env vars are only available when the PR is made from a fork in the same repo.

DilumAluthge:
Once https://github.com/JuliaRegistries/RegistryCI.jl/issues/122 is fixed, we’ll be able to run the integration tests on PRs from forks.

DilumAluthge:
But anyway, this PR looks fine to me. Fredrik or anyone feel free to YOLO merge. Or if no one objects after another day or so, I’ll YOLO merge.

DilumAluthge:
@timholy I have deployed this into production. Do you have any Gtk etc. packages to register so that we can see if it works?

DilumAluthge:
Also, it doesn't seem to be causing trouble with any other (e.g. non-GUI) packages.

So all that remains to be seen is if this fixes Gtk and other GUI packages.

timholy:
Yes, I have a pileup of these. But I first need to fix https://github.com/JuliaGraphics/Gtk.jl/pull/460, which is the same as https://github.com/JuliaGizmos/GtkReactive.jl/pull/91, upon which both ProfileView and ImageView depend.

DilumAluthge:
Just FYI. I added a debug statement to see if AutoMerge can find `xvfb-run`. The result:
```
┌ Debug: xvfb: 
│   xvfb = "/usr/bin/xvfb-run"
└ @ RegistryCI.AutoMerge ~/.julia/packages/RegistryCI/MU6OD/src/AutoMerge/guidelines.jl:179
```

So it looks like `xvfb-run` is in the `PATH` on GitHub Actions.

timholy:
It works! https://github.com/JuliaRegistries/General/pull/6407

PR #144, opened 2019-12-23
Add label when automerge fails

ericphanson:
I tried setting the automerge bot to add a label when the guidelines aren't met. I'm not sure how to test this locally without setting up a whole registry etc, so I thought I'd open a PR for feedback and to hear how to move forward.


src/AutoMerge/github.jl
```diff
@@ -87,6 +87,35 @@ function _get_updated_pull_request(pull_request::GitHub.PullRequest; auth::GitHu
...
+    issue = GitHub.issue(pr_base_repo, pr_number; auth=auth)
+    if issue.labels === nothing
+        labels = String[]
+    else
+        labels = [ d["name"] for d in issue.labels ]
+    end
+    return issue, labels
+end
+
+function add_label!(pull_request::GitHub.PullRequest, label::String; auth::GitHub.Authorization)
```
ericphanson: Regarding the `!`: I tried to follow the example of `merge!`; it doesn't actually modify the `pull_request::GitHub.PullRequest ` object itself, but it does modify the pull request on GitHub.

src/AutoMerge/github.jl
```diff
@@ -87,6 +87,35 @@ function _get_updated_pull_request(pull_request::GitHub.PullRequest; auth::GitHu
...
+        push!(labels, label)
+        GitHub.edit_issue(base_repo(pull_request), issue; auth=auth, params = Dict("labels"=>labels))
+    end
+end
+
+function remove_label!(pull_request::GitHub.PullRequest, label::String; auth::GitHub.Authorization)
+    issue, labels = _issue_labels(pull_request; auth=auth)
+    if label ∈ labels
+        unique!(labels)
+        deleteat!(labels, findfirst(isequal(label), labels))
```
ericphanson: I'm not sure these two lines are the best way to do this. I went for `isequal(label)` over `==(label)` since `∈` and `unique` use `isequal`. I don't think it should matter here though.

src/AutoMerge/AutoMerge.jl
```diff
@@ -10,6 +10,8 @@ import JSON
 
 import ..RegistryCI
 
+const AUTOMERGE_FAILURE_LABEL = "automerge failed"
```
ericphanson: I considered "needs manual merge" here, but I think most of the time the automerge fails, the user will want to update the PR to obey the guidelines. One problem with "automerge failed" though is it sounds like something went wrong in the process (like an exception) instead of an intended possible outcome. Maybe "guidelines not met" or something?

DilumAluthge: I think "automerge failed" is a fine name.

src/AutoMerge/github.jl
DilumAluthge: Yeah in this codebase I used `!` as a reminder to myself that it modifies some GitHub object.

src/AutoMerge/github.jl
DilumAluthge: Probably simpler to just use `filter`. I pushed some changes.

DilumAluthge:
Can you rebase on master and fix the merge conflicts?

ericphanson:
I rebased on master and force pushed to this branch, but the diff looks like there's lots of other recent changes. Did I mess it up?

DilumAluthge:
So, I gave this some more thought. I have realized that when you are in the list of pull requests (https://github.com/JuliaRegistries/General/pulls), you can just look at the status icon to figure out whether or not automerge is passing. If it is a green check mark ✅ , then automerge passed. If it is a red ❌ , then automerge failed.

So I don't think we also need to add a label. There's no real benefit to the labels, because you can look at the icon. And the cost is that it is more API calls, which counts towards our rate limit and increases the risk of us exceeding our rate limit.

ericphanson:
Ah ok, that makes sense; I’ll close this then. Thanks for the review, by the way. It was interesting seeing how the automerge code works; maybe I can contribute something else in the future now that I’m slightly familiar with it.

DilumAluthge:
There are several open issues that you might be interested in looking at :) I'm happy to help out with reviewing more pull requests.

PR #181, opened 2019-12-30
Refactor integration tests and add JLL integration tests

ericphanson:
I also threw in a test for `difference` since I saw the error path wasn't hit in e.g. the integration tests.


DilumAluthge:
This looks great. I'll review it more thoroughly once I've merged #183.

PR #184, opened 2019-12-30
WIP: Better JLL enforcement

### LINKED ISSUE #179: Non-jll package must not be named with suffix `_jll`
I think we should add this check for new (regular) package registrations, since we use `endswith(name, "_jll")` to check if a dependency of a JLL package is a JLL package. Also it's just confusing if someone could register a regular package that way.

DilumAluthge:
Fixes #179 

cc: @ericphanson


DilumAluthge:
I think we should have two types of authorized authors. One type is allowed to merge JLL but not allowed to merge non-JLL. Example is @jlbuild. The other type is allowed to merge non-JLL but not allowed to merge JLL. Example is @JuliaRegistrator.

PR #118, opened 2019-11-30
allow loading external registry deps

colinxs:
This allows the UUIDs tested to be loaded from external registries instead of just the registry being tested.

PR #194, opened 2020-01-05
Permit specifying dependency registries

BenJWard:
Closes #118 

This is a redux of #118.
I've used Pkg.Registry.add as suggested by @fredrikekre, which also allowed me to use Pkg, to find and get the UUIDs from the Registry.toml files, without assuming where the registries were stored.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>"]
-version = "0.12.0"
+version = "0.11.2"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>"]
-version = "0.11.1"
+version = "0.12.0"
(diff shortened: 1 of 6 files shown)


BenJWard:
I've set the next version as 0.12 as to my mind this count's as a (v. minor) new feature.

DilumAluthge:
For pre-1.0, the rule is we bump minor if it is breaking, patch if it is nonbreaking.

BenJWard:
Ok cool I'll change it!

BenJWard:
Are the AutoMerge test failing because of something I've done?

DilumAluthge:
Nah they always fail on forks. I’ll run them manually before I merge.

DilumAluthge:
Would be nice to have some tests.

BenJWard:
Sure I could add a test.

Although General is still used for most packages and only the newest BioJulia offerings are being released on the BioJuliaRegistry for now. Nevertheless, the BioJuliaRegistry is the only concrete example of a registry that I know of that has packages with dependencies that live in another registry (General), so I could use BioJuliaRegistry as a test case if you're happy with that.

DilumAluthge:
Yeah I think it’s fine to use BioJuliaRegistry in the tests

BenJWard:
Ok cool, I run the RegistryCI.test on BioJuliaRegistry in the same way it's run on General in the registryci_registry_testing.jl file. I'm not sure if you require any additional testing stuff - I saw some template PRs in a folder in there and stuff...

DilumAluthge:
@BenJWard The tests are currently failing because of an invalid compat entry in `BioJuliaRegistry`.

This pull request fixes that compat entry: https://github.com/BioJulia/BioJuliaRegistry/pull/28

BenJWard:
Merged it into BioJuliaRegistry. We were aware of the issue already - this issue was actually why I tried to get RegistryCI running on the registry over there, as we couldn't work out when that error got introduced as we have a Registrator bot make the PRs. So I thought, we need RegistryCI to stop this in the future 😆

PR #215, opened 2020-02-19
Drop the "can’t narrow Julia compat in a patch release" requirement for pre-1.0 packages

### LINKED ISSUE #204: Proposal: Drop the "can’t narrow Julia compat in a patch release" requirement for pre-1.0 packages
Currently, we don’t allow you to narrow the Julia compat in a patch release.

However, this means that if your package is pre-1.0, you have to make a breaking release in order to narrow the Julia compat, even if your package does not make any breaking changes.

I’d like to propose that we:
- Drop this requirement for pre-1.0 packages
- Keep this requirement for post-1.0 packages

DilumAluthge:
Fixes #204

### DIFF
--- a/src/AutoMerge/guidelines.jl
+++ b/src/AutoMerge/guidelines.jl
@@ -90,7 +90,7 @@ function meets_patch_release_does_not_narrow_julia_compat(pkg::String,
     if meets_this_guideline
         return true, ""
     else
-        if (old_version >= v"1") | (new_version >= v"1")
+        if (old_version >= v"1") || (new_version >= v"1")
             msg = string("A patch release is not allowed to narrow the ",
                          "supported ranges of Julia versions. ",
                          "The ranges have changed from ",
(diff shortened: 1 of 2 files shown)


DilumAluthge:
I’ll leave this open for a little while to allow for feedback.

src/AutoMerge/guidelines.jl
```diff
@@ -90,14 +90,19 @@ function meets_patch_release_does_not_narrow_julia_compat(pkg::String,
...
     else
-        msg = string("A patch release is not allowed to narrow the ",
-                     "supported ranges of Julia versions. ",
-                     "The ranges have changed from ",
-                     "$(julia_compats_for_old_version) ",
-                     "(in $(old_version)) ",
-                     "to $(julia_compats_for_new_version) ",
-                     "(in $(new_version)).")
-        return false, msg
+        if (old_version >= v"1") | (new_version >= v"1")
```
KristofferC: `||` is typically used unless you are actually bit twiddling.

DilumAluthge: Done!

PR #224, opened 2020-03-23
Make sure Deps.toml and Compat.toml roundtrips compression.

fredrikekre:


fredrikekre:
Fails due to changes like https://github.com/JuliaRegistries/General/pull/10324

DilumAluthge:
Hmmm. I feel like this is a good feature to have. Could we maybe hard code exceptions for the packages that are also in BioJuliaRegistry?

@kristofferc @stefankarpinski

src/registry_testing.jl
```diff
@@ -100,6 +107,13 @@ function test(path = pwd();
                 # Test that the way Pkg loads this data works
                 Test.@test Pkg.Operations.load_package_data_raw(Pkg.Types.VersionSpec, compatfile) isa
                     Dict{Pkg.Types.VersionRange,Dict{String,Pkg.Types.VersionSpec}}
+                # Make sure the content roundtrips through decompression/compression
+                compressed = RegistryTools.Compress.compress(compatfile,
+                    RegistryTools.Compress.load(compatfile))
+                if compressed != compat
```
DilumAluthge:
```suggestion
                if compressed != compat && !(compatfile in compatfile_exceptions)
```

DilumAluthge: And then you would define `compatfile_exceptions` as a constant somewhere in your code.

fredrikekre:
> Could we maybe manually add exceptions for the packages that are also in BioJuliaRegistry?

We should just fix the compression IMO.

KristofferC:
Regarding the BioJulia packages, I always felt it was weird that the compression made claims about versions that it does not yet know about. Having packages in multiple registries shows that that just isn't a good idea.

fredrikekre:
Yea, I think all ranges in Deps.toml and Compat.toml should be lower bounded to the lowest version in the current registry, and upper bounded on the highest version.

Of course, you can always sneak in a patch release in some other registry, but at least we have done what we can in that case.

fredrikekre:
So for example (https://github.com/JuliaRegistries/General/blob/7c811e09d6a174174e8f16d9cbd1a26651c66ad1/L/Literate/Compat.toml):
```toml
[0-1]
JSON = "0"
julia = ["0.7", "1"]

[2]
JSON = "0.18-0.21"
julia = "1"
```
should be
```toml
[0.3.0-1.1.0]
JSON = "0"
julia = ["0.7", "1"]

[2.0.0-2.3.0]
JSON = "0.18-0.21"
julia = "1"
```

fredrikekre:
Actually, maybe it is more important to fix Pkg (https://github.com/JuliaLang/Pkg.jl/issues/1434) and then it doesn't matter that a registry range cover more versions than what exist in the registry itself.

StefanKarpinski:
The intention was always that the compression was only with respect to the versions in _this_ registry. Any other claims that the compression appears to make are incidental, although since it follows semver logic, they should be somewhat reasonable. But yes, versions in other registries should not change the compression logic.

DilumAluthge:
How difficult will it be to fix https://github.com/JuliaLang/Pkg.jl/issues/1434?

DilumAluthge:
Looks like @KristofferC fixed https://github.com/JuliaLang/Pkg.jl/issues/1434 in https://github.com/JuliaLang/Pkg.jl/pull/1750

DilumAluthge:
@fredrikekre You'll probably want to rebase this on the latest master.

fredrikekre:
Good to go now.

DilumAluthge:
bors try

fredrikekre:
Looks like the test registry doesn't roundtrip :) At least we know the functionality works

DilumAluthge:
A great sanity check actually.

It shouldn't be too hard to fix the example registries . I'll take a look now

DilumAluthge:
Okay I got most of them. Let's see if that is sufficient.

bors try

fredrikekre:
You can use
```
#!/usr/bin/env julia

import RegistryTools.Compress
import Pkg.TOML

reg = TOML.parsefile(abspath(ARGS[1], "Registry.toml"))

for (uuid, data) in reg["packages"]
    for f in abspath.(ARGS[1], data["path"], ("Deps.toml", "Compat.toml"))
        if isfile(f)
            Compress.save(f, Compress.load(f))
        end
    end
end
```

DilumAluthge:
Cool, if Bors fails again, I'll run that script.

DilumAluthge:
Okay, I ran that script on all of the test registries.

@fredrikekre, should we also run that script on the General registry?

bors try

fredrikekre:
> @fredrikekre, should we also run that script on the General registry?

This PR does that in tests, and since they pass it's all good (fixed by https://github.com/JuliaRegistries/General/pull/20439).

DilumAluthge:
Tests should pass now.

bors merge

DilumAluthge:
Registration pending: https://github.com/JuliaRegistries/General/pull/20517

iamed2:
@fredrikekre Thanks for the fix script, we were able to fix the Invenia registry very quickly!

PR #226, opened 2020-04-06
Added note on mandatory waiting period length

logankilpatrick:
Added note on how long the mandatory waiting period is.


DilumAluthge:
It is already documented in the README here: https://github.com/JuliaRegistries/General#automatic-merging-of-pull-requests

logankilpatrick:
Right, but why not have it be user facing in the message? You are already saying that they have to wait, it doesn’t make sense to say okay now go read the docs to find out how long you have to wait.

DilumAluthge:
My only concern is that the more places we put the waiting period, that means there are more places that we need to change if we ever decide to change the length of the waiting period.

Anyway, since this is a new feature, can you bump the minor version number? And then let’s see what other people think.

logankilpatrick:
@DilumAluthge Conflicts have been resolved and version has been updated!

PR #241, opened 2020-04-18
Remove the dependency on GitCommand

DilumAluthge:
Git_jll requires a lot of artifacts.

It is getting really annoying to have to download those every time a Travis build needs to use RegistryCI.


DilumAluthge:
bors merge

PR #243, opened 2020-04-29
Disable the "URL ends in /name.jl.git" check

DilumAluthge:
We used to require that the repo URL ended in `/PackageName.jl.git` for new package registrations.

But now we support having multiple registered packages inside a single Git repo.

So this repo URL check no longer makes sense,

This PR disables that check.

cc: @StefanKarpinski

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>"]
-version = "2.2.1"
+version = "2.3.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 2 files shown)


DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
bors try-

DilumAluthge:
bors try

DilumAluthge:
Bors merge

StefanKarpinski:
Should this check remain in place if there's no subdirectory?

DilumAluthge:
> Should this check remain in place if there's no subdirectory?

Seems like a good idea.

We'll have to figure out a way for automerge to detect whether or not the package is in a sub directory. Maybe just open Package.toml and see if it contains the string "subdir"?

StefanKarpinski:
@GunnarFarneback: what is the appropriate criterion for this?

GunnarFarneback:
If I understand correctly what information is available, checking whether `Package.toml` has a 'subdir` field is indeed the way to go.

fredrikekre:
If this is only checked on new package registrations I think it is fine to just keep it as before an manually merge in those cases. I think it is nice to encourage the `.jl.git` pattern.

DilumAluthge:
> If this is only checked on new package registrations I think it is fine to just keep it as before an manually merge in those cases. I think it is nice to encourage the `.jl.git` pattern.

Seems fine to me!

PR #246, opened 2020-05-09
Allow new package names to end with uppercase

oxinabox:
This follows up to #200

Rather than just making it print a message that "Its OK to end with uppercase"
this just changes it to allow it.
I don't get what the motivation of #200 was, TBH.

The requirement to not end in uppercase is not listed in the summary in 
https://github.com/JuliaRegistries/General/blob/master/README.md#automatic-merging-of-pull-requests
which says
> Package name: Should start with a capital letter, contain only ASCII alphanumeric characters, and be at least 5 characters long.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>"]
-version = "2.3.0"
+version = "2.4.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Note that commenting on a pull request will automatically disable automerging on
 
 1. Normal capitalization
 
-    The package name should match `r"^[A-Z]\w*[a-z][0-9]?$"`, i.e. start with a capital letter, contain ASCII alphanumerics only, end in lowercase.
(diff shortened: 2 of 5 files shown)


oxinabox:
It seemed bad that the follow up to **PIML** (https://github.com/JuliaRegistries/General/pull/14364)
to be what is generally I think a good name **PhysicsInformedML** (https://github.com/JuliaRegistries/General/pull/14466) still can't be automerged.

DilumAluthge:
The goal of this check is to prevent names like `FOOBAR`.

What's wrong with manually merging names like `PhysicsInformedML`?

oxinabox:
> The goal of this check is to prevent names like FOOBAR.

Look at the tests (or the code).
It still prevents `FOOBAR`, but it now allows: `FOOiBAR` and `PhysicsInformedML`.
If someone does use this to do something aweful, well it doesn't really change much
they also can do `FOOBARi` which is no better.
And why we have the 3 day thing still.

> What's wrong with manually merging names like PhysicsInformedML?

Why do manually, when you can automate?

Also it is a bit of a off thing to have a note saying "What you are doing is allowed, but needs manually merging like things that are not allowed."
and makes it seem like it is not allowed.
Its a worse user experience,
particularly for exactly the case mentioned above,
where someone had a pretty unclear acronym, and then after Registry CI "computer said no"
was recommended to expend that acronym partially, only to have "computer still says no, but"

In general I think a lot of good package names end with common acronyms.
Often it is a adjective + an acronym. `MicroCSV`, `LazyJSON`, `LightXML`.

DilumAluthge:
Now I understand. This seems fine to me. Since @StefanKarpinski originally wrote this guideline, it would be good to get his input.

DilumAluthge:
In the meantime let's make sure the tests are passing.

bors try

oxinabox:
Bump

DilumAluthge:
Bors merge

PR #250, opened 2020-05-28
Improving usability for private registry, outside of github.com

### LINKED ISSUE #247: Missing way to provide URL of self-hosted github for AutoMerge.
So far the default url for github.com is used in `AutoMerge.run`, it would be great if there was option to specify it. 
Once https://github.com/JuliaWeb/GitHub.jl/pull/163 will be accepted, it would quite help with this issue.
Should I make a PR for this? (ideally after the abovementioned PR is merged and GitHub.jl will have new version)
Also, I wonder how the implementation should look like. Naive solution is to pass `api` parameter everywhere where `GitHub.jl` is called, which is lots of places. Or I might do something similar that `GitHub.jl` itself is using, with the `@api_default` macro https://github.com/JuliaWeb/GitHub.jl/blob/master/src/utils/requests.jl#L22 but here, the default API would not be constant, and it could be set in the `AutoMerge.run` argument. But that would require adding macro to all the calls, and would require explicitly passing the api to `GitHub.jl`, so I guess the former option would be more straightforward.

### LINKED ISSUE #248: Missing support for TeamCity in AutoMerge.
https://github.com/JuliaRegistries/RegistryCI.jl/blob/master/src/AutoMerge/ciservice.jl contains implementation for Github Actions and Travis CI. It would be great to have also support for TeamCity.

Should I make a PR for this? I'm wondering how it should be implemented though, not sure what the best practise would be. 
Travis CI and GitHub Actions are quite straightforward compared to TeamCity, which requires [Pull Requests build feature](https://www.jetbrains.com/help/teamcity/pull-requests.html) turned on, and also needs predefined build parameters to be passed as environment variables, because they are not present by default.
Should the also be Kotlin DSL included, which will configure necessary things? Then the code would depend on names of variables defined in that DSL. I find it unnecessarily complicated, but I haven't figured out any better solution.

### LINKED ISSUE #249: RegistryCI doesn't respect package server
I would like to be able to set "General" as dependent registry, without need to provide full URL, hoping Pkg would figure out which source to use base on either presence or absence of JULIA_PKG_SERVER, specifically in
https://github.com/JuliaRegistries/RegistryCI.jl/blob/master/src/AutoMerge/guidelines.jl#L231
https://github.com/JuliaRegistries/RegistryCI.jl/blob/master/src/AutoMerge/guidelines.jl#L271
https://github.com/JuliaRegistries/RegistryCI.jl/blob/aa08b2d1d4036dfdbd6c5373881b93c2ae6cd023/src/registry_testing.jl#L17

racinmat:
Fixes #247 
Fixes #248 
Fixes #249 

I'll keep it as drawn until https://github.com/JuliaWeb/GitHub.jl/pull/163 is merged because it is required for this to work properly.
Any feedback welcome. I hope this will make private registry more accessible by enabling such great tooling as is RegistryCI to support passing dependency registry by name without need for full URL, and allowing to parametrize the base API URI.

Later, I plan to add also support for TeamCity.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>"]
-version = "2.4.0"
+version = "2.5.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
@@ -15,7 +15,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 TimeZones = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 
 [compat]
-GitHub = "5.1.1"
+GitHub = "5.1.6"
 HTTP = "0.8"
 JSON = "0.19, 0.20, 0.21"
 RegistryTools = "1.2"
--- a/README.md
+++ b/README.md
@@ -54,6 +54,12 @@ Note that commenting on a pull request will automatically disable automerging on
 
 For instructions on how to run the RegistryCI.jl integration tests on your local machine, see [`INTEGRATION_TESTS.md`](INTEGRATION_TESTS.md).
 
+## TeamCity support
+
+There is support also for TeamCity, but it does not work out-of the box, it requires Pull Request build feature to be added, and passing few build variables as environment variables.
+
+To make it work in TeamCity, see [this Kotlin DSL snippet](teamcity_settings.kts) and put following parts into your DSL/add them in GUI.
+
 ## Acknowledgements
 
 Dilum Aluthge would like to acknowledge the following:
--- a/bin/remember_to_update_registryci.jl
+++ b/bin/remember_to_update_registryci.jl
@@ -15,7 +15,8 @@ end
     return nothing
 end
 
-function get_all_pull_requests(repo::GitHub.Repo,
+function get_all_pull_requests(api::GitHub.GitHubAPI,
+                               repo::GitHub.Repo,
                                state::String;
                                auth::GitHub.Authorization,
                                per_page::Integer = 100,
@@ -24,13 +25,13 @@ function get_all_pull_requests(repo::GitHub.Repo,
     myparams = Dict("state" => state,
                     "per_page" => per_page,
                     "page" => 1)
-    prs, page_data = GitHub.pull_requests(repo;
+    prs, page_data = GitHub.pull_requests(api, repo;
                                           auth=auth,
                                           params = myparams,
                                           page_limit = page_limit)
     append!(all_pull_requests, prs)
     while haskey(page_data, "next")
-        prs, page_data = GitHub.pull_requests(repo;
+        prs, page_data = GitHub.pull_requests(api, repo;
                                               auth=auth,
                                               page_limit = page_limit,
                                               start_page = page_data["next"])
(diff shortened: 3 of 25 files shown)


DilumAluthge:
Since this is a new feature, can you bump the minor version number in Project.tomo?

racinmat:
You're right, I'll bump the minor version, I just like to wait with modifying versions till the end because I'm not sure how many things will change before it will be merged.
But when it will be polished, ready and working with new version, I'll definitely bump it.

DilumAluthge:
> You're right, I'll bump the minor version, I just like to wait with modifying versions till the end because I'm not sure how many things will change before it will be merged.
> But when it will be polished, ready and working with new version, I'll definitely bump it.

Sounds great!

racinmat:
Hopefully now everything should be OK, I bumped version and compat, I added note about TeamCity to readme, 
I hope using HTTP package to validate the URL in https://github.com/JuliaRegistries/RegistryCI.jl/pull/250/files#diff-6a40dfcffc694ae828b773fda837b1f0R229 and https://github.com/JuliaRegistries/RegistryCI.jl/pull/250/files#diff-6a40dfcffc694ae828b773fda837b1f0R276 won't be causing any problem once it will be doing this for the HTTP package itself, but in that case, I could always put there hardcoded url validation without using any packages.

racinmat:
Is there some reason, why Travis CI and bors did not start automatically, @DilumAluthge ?

DilumAluthge:
bors try

racinmat:
I see that https://github.com/BioJulia/BioJuliaRegistry has been deprecated and is merged into General registry. 
And since it's used in tests here https://github.com/racinmat/RegistryCI.jl/blob/master/test/registryci_registry_testing.jl#L18 I'm not sure if there is some other public registry to test if  handling dependent registries works or not.

Should we look for some more registries, or just delete this test which uses it?

racinmat:
bors try

racinmat:
@DilumAluthge  I removed the test, but I need someone to run the bors.

DilumAluthge:
bors try

racinmat:
I have no idea what went wrong here, I just noticed that for the test package, https://github.com/MikeInnes/Requires.jl.git is used, although the newer one is https://github.com/JuliaPackaging/Requires.jl.git and that tests use
```
["1.0.0"]
git-tree-sha1 = "f6fbf4ba64d295e146e49e021207993b6b48c7d1"
```
although in https://github.com/JuliaRegistries/General/blob/master/R/Requires/Versions.toml there is 
```
["0.5.2"]
git-tree-sha1 = "f6fbf4ba64d295e146e49e021207993b6b48c7d1"

["1.0.0"]
git-tree-sha1 = "999513b7dea8ac17359ed50ae8ea089e4464e35e"
```

DilumAluthge:
Try changing the "f6fbf4ba64d295e146e49e021207993b6b48c7d1" hash to "999513b7dea8ac17359ed50ae8ea089e4464e35e" and see if it fixes things.

racinmat:
done

DilumAluthge:
Can you rebase on master, and squash?

racinmat:
squashed

DilumAluthge:
bors try

racinmat:
Cool, it passes. Coverage decreased, but I guess this is to be expected (although it's unfortunate).

DilumAluthge:
bors merge

DilumAluthge:
bors merge

DilumAluthge:
bors try

DilumAluthge:
Test failures now are due to https://github.com/JuliaLang/Pkg.jl/pull/1841

I think I've fixed it now though.

DilumAluthge:
bors merge

PR #255, opened 2020-06-17
Make sure all paths are unique.

fredrikekre:


DilumAluthge:
@fredrikekre Can you bump the version number based on whether you think this is a new feature or a bugfix?

DilumAluthge:
@fredrikekre Just to confirm, this makes sure that all paths are unique even on case-insensitive filesystems, right?

bors merge

PR #265, opened 2020-08-26
Reenable and fix tests that check Pkg can read the entries

fredrikekre:
Reenable and fix tests that check Pkg can read and understand compat and deps entries in the registry.

This thus also reverts commit 9c6f8fc2ce7393695c9040cb0ec10bd1adc1ba5f.


DilumAluthge:
bors merge

PR #268, opened 2020-09-01
Give each tested package in a registry a named testset

iamed2:
When tests failed here I was not able to tell from the CI logs which packages were failing. Now this is possible!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>"]
-version = "2.8.0"
+version = "2.9.0"
 
 [deps]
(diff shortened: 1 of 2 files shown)


fredrikekre:
Does it affect the timing at all from e.g. testing the General registry? I know that sometimes nesting testsets takes some extra time.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>"]
-version = "2.8.0"
+version = "2.8.1"
```
fredrikekre, line 4: Maybe 2.9?

fredrikekre (approved):

iamed2:
> Does it affect the timing at all from e.g. testing the General registry? I know that sometimes nesting testsets takes some extra time.

Not significantly, both take between 4 and 5 seconds on my computer.

Project.toml
iamed2: Sure

fredrikekre:
bors merge

PR #270, opened 2020-09-22
Allow Artifacts and JLLWrappers as deps of JLLs

giordano:
New jll packages are going to have two additional dependencies: `Artifacts` (a new stdlib in Julia v1.6, [this package](https://github.com/JuliaPackaging/Artifacts.jl) for the other versions) and [`JLLWrappers`](https://github.com/JuliaPackaging/JLLWrappers.jl).

Automerge in https://github.com/JuliaRegistries/General/pull/21815 also complains about `Artifacts.toml` not having a compat bound.  Not sure what to with that one, since it's a new stdlib, maybe we should set [v1.3](https://github.com/JuliaPackaging/Artifacts.jl/blob/258eaa57c9ec0d488c2082743644cd03b7691408/Project.toml#L3) as minimum version?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>"]
-version = "3.0.0"
+version = "3.0.1"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/AutoMerge/jll.jl
+++ b/src/AutoMerge/jll.jl
@@ -21,16 +21,18 @@ end
 function meets_allowed_jll_nonrecursive_dependencies(working_directory::AbstractString,
                                                      pkg,
                                                      version)
-    # If you are a JLL package, you are only allowed to have three kinds of dependencies:
+    # If you are a JLL package, you are only allowed to have five kinds of dependencies:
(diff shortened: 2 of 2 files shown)


vchuravy:
Uhm... how are we dealing with backwards compatibility? E.g. running these JLLs on 1.5, where Artifacts doesn't exist yet?

giordano:
That's what [`Artifacts.jl`](https://github.com/JuliaPackaging/Artifacts.jl) should be for

vchuravy:
So we do need a compat entry for that.

DilumAluthge:
1. Can you bump the version number? I guess this is a bug fix, right?
2. Regarding the Artifacts stdlib/Artifacts.jl: let's just special case it and exempt it from the compat requirements, since in 1.6+ it's a stdlib anyway. We can do this in a separate PR.

giordano:
Well, wrong commit message, right commit content

DilumAluthge (approved):
LGTM

DilumAluthge:
bors merge

DilumAluthge:
@giordano After this is merged, you'll need to register it with JuliaRegistrator.

Once the new version of RegistryCI is registered, you can update the manifest in the General registry (in the `.CI`) directory so that the manifest uses the latest version of RegistryCI. Of course, you can do this manually by cloning the General registry, instantiating, and updating. Or you can do it automatically. To do so, go to the General registry, click on the Actions tab, in the list of Actions click on "remember_to_update_registry_ci", then click the green "run action" button. It will open a PR to update the manifest.

Or you can just wait an hour, and it will automatically open a PR to update the manifest.

PR #272, opened 2020-09-25
Make the AutoMerge comments more concise

DilumAluthge:
The AutoMerge comments have gotten way too long and wordy.

If a package does not meet the AutoMerge guidelines, I think it is very reasonable to expect that the package's author will read the General registry README. We don't need to repeat all of that information in each AutoMerge comment.


DilumAluthge:
bors merge

PR #275, opened 2020-10-10
Disable the "standard initial version number" check

DilumAluthge:
Of all of the AutoMerge checks, this is the only one that I don't think is useful. In 100% of the cases in which a package fails this check, we always manually merge it anyway.

If it's not changing what we do, there's no need to have the check in the first place.


DilumAluthge:
Bors merge

PR #277, opened 2020-10-12
Don't repeat documentation that is available elsewhere

DilumAluthge:
There's no point in having the same documentation repeated in multiple places.

I think we should just maintain this list in the General registry README. Then, other locations can just link to the General registry README.


DilumAluthge:
@ericphanson Can you make a PR to add the visual distance, etc. docs to the General registry README?

ericphanson:
Sure, one sec

ericphanson (approved):

DilumAluthge:
Bors merge

PR #281, opened 2020-10-17
Check name distance to standard libraries

ericphanson:
As came up in https://github.com/JuliaRegistries/General/pull/23128

I first thought to do this by adding the stdlibs to the `all_pkg_names` variable here https://github.com/JuliaRegistries/RegistryCI.jl/blob/e4fff963024a5091b4d852fbcdab7284aab18db5/src/AutoMerge/new-package.jl#L140

But then I remembered I had added to the [General registry's docs](https://github.com/JuliaRegistries/General#automatic-merging-of-pull-requests) the following:

> To test yourself that a tentative package name, say MyPackage meets these checks, you can use the following code (after adding the RegistryCI package to your Julia environment):
> ```julia
> using RegistryCI
> using RegistryCI.AutoMerge
> all_pkg_names = AutoMerge.get_all_non_jll_package_names(path_to_registry)
> AutoMerge.meets_distance_check("MyPackage", all_pkg_names)
> ```
> where path_to_registry is a path to the folder containing the registry of interest. 

and I wanted to keep that working and keep it as simple as possible. So I decided to just add the stdlibs to the return value of `AutoMerge.get_all_non_jll_package_names`. I also added a docstring to that function to make it more obvious that it returns the stdlibs and not just packages from the registry.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>"]
-version = "4.1.0"
+version = "4.1.1"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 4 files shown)


ericphanson:
I bumped the version as a bugfix since stdlibs were omitted by mistake from the original PR.

DilumAluthge (approved):
LGTM

fredrikekre (approved):

PR #289, opened 2020-10-30
Use TagBot token for merging stuff

christopher-dG:
When Automerge merges PRs, it does not trigger the TagBot notifier because actions can't trigger other actions with the default token. So for cron jobs, use a user token.


DilumAluthge:
@christopher-dG Integration tests are failing.

christopher-dG:
bors merge

PR #290, opened 2020-10-30
Filter out pull requests when looking for a trigger issue

christopher-dG:
Slight oversight: When looking for a closed issue by JuliaTagBot, it also includes PRs. So it picks up the PR adding the TagBot action from way back when: https://github.com/JuliaRegistries/RegistryCI.jl/pull/214

PR #291, opened 2020-10-30
Fix the TagBot workflow file

DilumAluthge:


christopher-dG:
I did this in #290 actually

PR #303, opened 2020-11-21
The "new package name should not include `julia` in the name" check should apply to all packages (including JLL packages)

DilumAluthge:
If for some reason a new JLL package wants to include "julia" in its name, I think it's perfectly fine to require manual merge in that case.


DilumAluthge:
bors merge

ericphanson:
Hmm, my thinking was that they get manual review over at Yggdrasil either way, and there are Julia JLLs (https://github.com/JuliaPackaging/Yggdrasil/tree/master/J/Julia) so there are legit uses for it there. But since it’s a new-package-only check it shouldn’t come up too often anyway.

DilumAluthge:
> But since it’s a new-package-only check it shouldn’t come up too often anyway.

Yeah that was my thinking.

PR #314, opened 2020-12-29
Simplification of code using undef.

GunnarFarneback:
This PR removes all occurrences of `undef`. Not because there's anything wrong with `undef` but because all uses could be simplified. There should be no functional differences.

### DIFF
--- a/src/AutoMerge/github.jl
+++ b/src/AutoMerge/github.jl
@@ -66,7 +66,7 @@ function get_all_my_pull_request_comments(api::GitHub.GitHubAPI,
     all_comments = get_all_pull_request_comments(api, repo,
                                                  pr;
                                                  auth = auth)
-    my_comments = Vector{GitHub.Comment}(undef, 0)
+    my_comments = GitHub.Comment[]
     for c in all_comments
         if c.user.login == whoami
             push!(my_comments, c)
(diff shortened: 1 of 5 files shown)


src/AutoMerge/github.jl
```diff
@@ -112,12 +112,7 @@ end
 
 function get_changed_filenames(api::GitHub.GitHubAPI, repo::GitHub.Repo, pull_request::GitHub.PullRequest; auth::GitHub.Authorization)
     files = GitHub.pull_request_files(api, repo, pull_request; auth=auth)
-    n = length(files)
-    filenames = Vector{String}(undef, n)
-    for i = 1:n
-        filenames[i] = files[i].filename
-    end
-    return filenames
+    return [file.filename for file in files]
```
GunnarFarneback, line 115: Or `return getfield.(files, :filename)` if that would be preferred.

DilumAluthge:
(I'm requesting a review from @ericphanson because I figure having people besides myself review these PRs will help get more people familiar with the RegistryCI code base.)

src/AutoMerge/github.jl
ericphanson: I think we should stick with `getproperty` as the current code uses; for the comprehension vs broadcasting, I think the comprehension is slightly easier to understand, but either way seems OK.

ericphanson (approved):
Thanks @GunnarFarneback!

src/AutoMerge/github.jl
GunnarFarneback: Let's leave it as it is.

ericphanson:
bors try

ericphanson:
bors merge

PR #315, opened 2020-12-29
Simplify logic by unnesting if expressions.

GunnarFarneback:
This PR simplifies the implementation of `cron_or_api_build(api, pr, registry)` by repeatedly applying the rewrite from
```
if x
    y
else
    z
end
return nothing
```
to
```
if !x
    z
    return nothing
end
y
return nothing
```
which dramatically reduces the nesting depth of the code and makes the logic easier to follow. There should be no functional differences.

Review tip: The only effective way to review is probably to apply these transformations yourself and see if you end up with the same code.

### DIFF
--- a/src/AutoMerge/cron.jl
+++ b/src/AutoMerge/cron.jl
@@ -271,152 +271,158 @@ function cron_or_api_build(api::GitHub.GitHubAPI,
     pr_number = number(pr)
     @info("Now examining pull request $(pr_number)")
     pr_author = author_login(pr)
-    if pr_author in vcat(authorized_authors, authorized_authors_special_jll_exceptions)
-        if is_new_package(pr) || is_new_version(pr)
-            if is_new_package(pr) # it is a new package
-                pr_type = :NewPackage
-                pkg, version = parse_pull_request_title(NewPackage(), pr)
-            else # it is a new version
-                pr_type = :NewVersion
-                pkg, version = parse_pull_request_title(NewVersion(), pr)
-            end
-            pr_age = time_since_pr_creation(pr)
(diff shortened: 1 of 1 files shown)


src/AutoMerge/cron.jl
```diff
@@ -271,152 +271,158 @@ function cron_or_api_build(api::GitHub.GitHubAPI,
...
-                                                                   all_statuses;
-                                                                   auth = auth)
-                    _checkruns_good = all_specified_check_runs_passed(api, registry,
-                                                                      pr,
-                                                                      passed_pr_head_sha,
-                                                                      all_check_runs;
-                                                                      auth = auth)
-                    if _statuses_good && _checkruns_good
-                        if pr_has_no_blocking_comments(api, registry, pr; auth = auth)
-                            "Pull request: $(pr_number). "
```
GunnarFarneback, line 319: Why are there loose strings here? Was this supposed to be an `@info`?

DilumAluthge: Yeah that should be an `@info`.

GunnarFarneback: Actually there's an `@info` with exactly those strings a few lines below, so these copies should probably just be deleted.

ericphanson:
Thanks @GunnarFarneback! I’ll review both of these in the next few days.

ericphanson (approved):
looks good! Much easier to follow, and it makes sense as sequence of checks to see if we should reject (and if so, to reject immediately).

ericphanson:
Let's check CI:

bors try

ericphanson:
bors merge

PR #316, opened 2020-12-30
Reduce nesting of if expressions.

GunnarFarneback:
Similar to #315 but on a smaller scale. In this case `git log -p -w` is good for reviewing.


ericphanson (approved):

ericphanson:
> In this case git log -p -w is good for reviewing

Thanks, that reminded me of the hide whitespace option in github (which adds the `&w=1` bit to the URL, like https://github.com/JuliaRegistries/RegistryCI.jl/pull/316/files?diff=unified&w=1), very helpful!

ericphanson:
bors merge

PR #318, opened 2021-01-03
Restructure guideline verification

GunnarFarneback:
~(Note: The first commit is #316. The PR will be rebased once #316 has been merged.)~

This PR adds some more structure to the guideline verification. ~For now it's more important to review the design ideas documented here than the partial implementation.~

### Step 1
The first commit introduces a `GitHubAutoMergeData` struct. This is a small gain by itself since it avoids passing around a dozen arguments to two functions and adds some documentation of the struct fields. However, the real purpose is to make it possible to uniformly organize the guideline checks as functions of `GitHubAutoMergeData` in preparation for the next step.

The second commit makes some normalizations to how the guideline checks are called, also in preparation for future steps.

### Step 2
Introduce a `Guideline` struct, that allows collecting all information about a guideline check, e.g.
```
struct Guideline
    info::String
    check::Function
    passed::Bool
    message::String
end

g1 = Guideline("Only modifies the files that it's allowed to modify",
               data -> pr_only_changes_allowed_files(data.api,
                                                     data.registration_type,
                                                     data.registry,
                                                     data.pr,
                                                     data.pkg;
                                                     auth = data.auth))
```
with methods to run the check and extract status and message.

This is implemented in commit 3 and then the guideline definitions are moved around and deduplicated in commit 4.

### Step 3
Replace the sequential code in the `pull_request_build` methods by collecting a vector of the applicable `Guideline` objects
depending on whether it is a new package or version and whether it is a jll or not, then just loop over that vector.

This is implemented in commit 6. Commit 5 consists of a couple minor fixes.


ericphanson:
this seems like a great approach to me. Should make adding or editing guidelines very easy.

src/AutoMerge/pull-requests.jl
```diff
@@ -95,51 +99,40 @@ function pull_request_build(api::GitHub.GitHubAPI,
...
         )
+        return nothing
+    end
+
+    registry_master = clone_repo(registry)
+    if !master_branch_is_default_branch
+        checkout_branch(registry_master, master_branch)
     end
+    pkg, version = parse_pull_request_title(registration_type, pr)
+    data = GitHubAutoMergeData(;api,
```
DilumAluthge, line 119: I'd like to keep compatibility with Julia 1.3. The syntax `foo(; bar, baz)` isn't available in Julia 1.3, so unfortunately you'll need to write it out as `foo(; bar = bar, baz = baz)`.

GunnarFarneback: I know (the PR bumps the compat) but if we're not ready to abandon 1.3 I'll change it. There's a lot of code in this package that would gain from that particular 1.5 feature.

DilumAluthge: Hmmmmmmmmmm.

DilumAluthge: Let's drop 1.3 and 1.4.

I know that people are running AutoMerge on private registries, but I think it's reasonable for them to upgrade AutoMerge to Julia 1.5 or 1.6 if they want new AutoMerge features.

Can you also edit the GitHub Actions workflow files so that the unit tests are run on:
- `'1.5'`
- `'1'`
- `'nightly'`

GunnarFarneback: Well, I changed it to 1.3 compatible. Let's leave it like that for now, then we can introduce that feature all over the package at once after this PR is finished.

DilumAluthge:
The test failures are no longer due to this PR. It's because I merged something into General master that requires the 1.6 beta.

DilumAluthge:
You'll need to rebase after https://github.com/JuliaRegistries/RegistryCI.jl/pull/337 is merged, and then CI should pass.

DilumAluthge:
Alright if you rebase and squash now, CI should pass.

src/AutoMerge/pull-requests.jl
DilumAluthge: Sounds good.

GunnarFarneback:
This is ready for review. When checking for correctness it's probably best to do it commit by commit.

Notes:
1. There are functional changes here. In particular non-applicable guidelines will no longer be checked (and then ignored), nor reported by name.
2. There are opportunities left from the last commit to further streamline the guideline checking and reduce the duplication of code between `new-package` and `new-version` but I think it's better to stop here. The main effect of the restructuring has been achieved and further improvements can be done in independent PRs without increasing the review burden of this one.

DilumAluthge:
>     1. There are functional changes here. In particular non-applicable guidelines will no longer be checked (and then ignored), nor reported by name.

Can you elaborate on this?

GunnarFarneback:
For example here, if we have a non-jll package, https://github.com/JuliaRegistries/RegistryCI.jl/blob/4cee83b64929b2339169a3fbc03044bcaa324810/src/AutoMerge/new-version.jl#L108-L117
the `meets_allowed_jll_nonrecursive_dependencies` computation will be performed and then ignored. This function call will not be made after this PR for non-jll packages.
Furthermore the corresponding `@info` print https://github.com/JuliaRegistries/RegistryCI.jl/blob/4cee83b64929b2339169a3fbc03044bcaa324810/src/AutoMerge/new-version.jl#L134-L136
will go away in the non-jll case.

DilumAluthge:
Oh I see. Yeah that is fine.

src/AutoMerge/new-package.jl
```diff
@@ -79,170 +79,70 @@ function pull_request_build(data::GitHubAutoMergeData, ::NewPackage)::Nothing
...
                       context = "automerge/decision",
                       description = "New package. Failed.")
     end
 
-    G12 = guideline_version_can_be_pkg_added
-    g12, m12 = check!(G12)
-    @info(G12.info,
-          meets_this_guideline = g12,
-          message = m12)
+    push!(guidelines,
```
ericphanson, line 134: I think it might be nicer to have two separate vectors, like `guidelines_stage_1` and `guidelines_stage_2`, where here we are starting `stage_2`. That would help that if we add a "stage 2" guideline, we don't have to ensure we update the indexing `guidelines[end-1:end]` below at the same time (since instead at the second check, we just loop over `guidelines_stage_2`. Then we could `vcat` them into just `guidelines` later if that's helpful.

ericphanson (commented):
The only correctness issue I could identify is that I think `Guideline` needs to be mutable (the integration tests would surely catch this too). Besides that I just have a few small comments.

I'll run CI too to double-check my mutability concern and to see if it sees anything else:

bors try

src/AutoMerge/pull-requests.jl
```diff
@@ -95,51 +99,40 @@ function pull_request_build(api::GitHub.GitHubAPI,
...
+    pkg, version = parse_pull_request_title(registration_type, pr)
+    data = GitHubAutoMergeData(;api = api,
+                               registration_type = registration_type,
+                               pr = pr,
+                               pkg = pkg,
+                               version = version,
+                               current_pr_head_commit_sha = current_pr_head_commit_sha,
+                               registry = registry,
+                               auth = auth,
+                               authorized_authors_authors = authorized_authors_authors,
```
ericphanson, line 127:
```suggestion
                               authorized_authors = authorized_authors,
```
found by the integration tests

ericphanson:
fixed a trivial bug so re-running:

bors try

ericphanson:
@GunnarFarneback, I'll give you bors permissions for this PR, so you can run integration tests yourself (via `bors try`).

bors delegate+

src/AutoMerge/jll.jl
```diff
@@ -1,3 +1,9 @@
+# This is a somewhat artificial guideline that always fails. The point
+# is the messages.
```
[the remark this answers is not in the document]
DilumAluthge: Good catch. Yes the JLL only authors are not allowed to register non-JLL packages.

src/AutoMerge/types.jl
```diff
@@ -31,3 +31,64 @@ end
...
+    # "General" when running automerge for a private registry.
+    registry_deps::Vector{String}
+end
+
+# Constructor that requires all fields as named arguments.
+function GitHubAutoMergeData(;kwargs...)
+    fields = fieldnames(GitHubAutoMergeData)
+    @assert Set(keys(kwargs)) == Set(fields)
+    return GitHubAutoMergeData(getindex.(Ref(kwargs), fields)...)
+end
```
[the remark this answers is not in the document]
GunnarFarneback: I've never been using that. Sounds like a possible followup PR.

src/AutoMerge/guidelines.jl
```diff
@@ -72,6 +72,10 @@ function meets_patch_release_does_not_narrow_julia_compat(pkg::String,
                                                           registry_head::String,
                                                           registry_master::String)
     old_version = latest_version(pkg, registry_master)
+    if old_version.major != new_version.major || old_version.minor != new_version.minor
+        # Not a patch release.
+        return true, ""
+    end
```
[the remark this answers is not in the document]
GunnarFarneback: Possibly, but only by accident if so. The real issue is that `old_version` should be computed as closest preceding version, not latest version. I think this is a more widespread issue.

src/AutoMerge/types.jl
```diff
@@ -92,3 +92,31 @@ function GitHubAutoMergeData(;kwargs...)
...
+    # This initial status and message are overwritten when the check
+    # has been run.
+    passed = false
+    message = "Internal error. A check that was supposed to run never did: $info"
+    return Guideline(info, check, passed, message)
+end
+
+passed(guideline::Guideline) = guideline.passed
+message(guideline::Guideline) = guideline.message
+function check!(guideline::Guideline, data::GitHubAutoMergeData)
```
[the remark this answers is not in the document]
GunnarFarneback: Yeah, this can't possibly work. Not sure what I was thinking. Probably didn't at all.

src/AutoMerge/jll.jl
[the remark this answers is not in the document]
GunnarFarneback: The check is fine and has been there all the time. What's artificial is that it's not actually checked from the `Guideline` object but from the outside code.

src/AutoMerge/new-package.jl
```diff
@@ -1,3 +1,8 @@
+# TODO: Probably want to get rid of this pseudo-guideline.
+const guideline_unimplemented =
+    Guideline("TODO: implement this check",
+              data -> (true, ""))
```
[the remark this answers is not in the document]
GunnarFarneback: Yes, this is part of note 2.

src/AutoMerge/new-package.jl
GunnarFarneback: Agreed that this isn't so nice and also part of note 2. My idea was to have a single input queue but inject `:update_status` or something to that effect between the guidelines, when it's time to update the status. (If that's needed at all, I'm not clear  on the timing aspects of these checks.)

src/AutoMerge/types.jl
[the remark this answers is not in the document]
ericphanson: sure, sounds good

src/AutoMerge/pull-requests.jl
GunnarFarneback: That's a weird mistake. (Well, the right hand side was done by a keyboard macro, so the original typo was only made once.)

src/AutoMerge/jll.jl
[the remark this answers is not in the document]
ericphanson: ah I see, yeah. It would be nice to move that checking logic into the guideline, but that sounds like a followup item as well

GunnarFarneback:
bors try

src/AutoMerge/new-package.jl
DilumAluthge: You can think of the checks in two parts: 
1. Everything before part 2
2. Part 2, which is `Pkg.add` and `import`.

The part 2 checks can take a long time depending on the package. If the part 1 checks fail, then we update the AutoMerge/decision status with a fail before we run part 2, so that people can get quick feedback even while they wait for the part 2 checks to finish.

GunnarFarneback:
bors try

GunnarFarneback:
bors try

GunnarFarneback:
bors try

GunnarFarneback:
bors try

ericphanson:
Thanks @GunnarFarneback!

bors merge

PR #320, opened 2021-01-10
Migrate from Travis to GitHub Actions

DilumAluthge:


ericphanson:
what was breaking here? I'm looking at updating our RegistryCI but having trouble seeing what the breaking changes were

DilumAluthge:
🤦 

I don't see any breaking changes in this PR.

I think I was planning on making a couple of breaking changes before the next release, which is why I bumped the version number as such (to satisfy VersionVigilante). But it doesn't look like that happened.

ericphanson:
Ok no worries, thanks for the confirmation! I'll feel safer updating then 🙂

PR #324, opened 2021-01-10
AutoMerge: If the PR is not from `JuliaRegistrator` or `jlbuild`, exit with success

DilumAluthge:
People frequently get confused when they open a manual PR to General, and then the AutoMerge job "fails".

So, on those PRs, instead of exiting with a failure ❌, let's exit with a success ✅.

IMPORTANT: just because the AutoMerge job "succeeds", doesn't mean that AutoMerge has passed on those PRs. It just means that AutoMerge skipped those PRs. But GitHub doesn't have a "skip" option for a run; either a run succeeds, or it fails.


DilumAluthge:
BTW, this behavior is easy to change. Just provide the `error_exit_if_automerge_not_applicable` keyword argument to the `AutoMerge.run` function.
- `error_exit_if_automerge_not_applicable = false`: If the PR is not from `JuliaRegistrator` or `jlbuild`, exit with success
- - `error_exit_if_automerge_not_applicable = true`: If the PR is not from `JuliaRegistrator` or `jlbuild`, exit with error (failure)

The default value is `error_exit_if_automerge_not_applicable = false`.

DilumAluthge:
bors merge

DilumAluthge:
bors merge

PR #328, opened 2021-01-10
Only run the integration tests on the `staging` and `trying` branches

DilumAluthge:
There's not really any point in running the integration tests on the `master` branch. After all, we use Bors. And Bors only fast-forwards `master` to point to `staging` if the integration tests have already passed on `staging`. So it's redundant to run them again on `master`, since we know that they have already passed on the exact same commit that `master` now points to.


DilumAluthge:
bors merge

PR #337, opened 2021-01-11
Fix CI

DilumAluthge:
The General registry now has a version of a package (CUDA 2.5) that relies on some Julia 1.6-specific stuff. So tests are expected to fail on Julia 1.5 and earlier.


DilumAluthge:
bors try

DilumAluthge:
bors merge

PR #341, opened 2021-01-16
Nicer collection and processing of guidelines.

GunnarFarneback:
Followup to #318, making the collection and processing of guidelines a lot nicer.

### DIFF
--- a/src/AutoMerge/new-version.jl
+++ b/src/AutoMerge/new-version.jl
@@ -47,6 +47,10 @@ function pull_request_build(data::GitHubAutoMergeData, ::NewVersion)::Nothing
         this_pr_can_use_special_jll_exceptions = false
     end
 
+    # If this is true it means that the author only is authorized for
+    # jll packages but this is is a normal package.
+    # TODO: Do all authorization checks in one place before calling
+    # this function.
     jll_only_authorization = (!this_is_jll_package
                               && pr_author_login ∉ data.authorized_authors)
 
--- a/src/AutoMerge/new-version.jl
+++ b/src/AutoMerge/new-version.jl
@@ -27,8 +27,9 @@ function pull_request_build(data::GitHubAutoMergeData, ::NewVersion)::Nothing
     #     - i.e. can we run `import Foo`
     pr_author_login = author_login(data.pr)
     this_is_jll_package = is_jll_name(data.pkg)
-    pkg, version = data.pkg, data.version
-    @info("This is a new package pull request", pkg, version,
+    @info("This is a new package pull request",
+          pkg = data.pkg,
+          version = data.version,
           this_is_jll_package)
 
     update_status(data;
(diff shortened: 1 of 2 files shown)


src/AutoMerge/new-version.jl
```diff
@@ -46,57 +47,49 @@ function pull_request_build(data::GitHubAutoMergeData, ::NewVersion)::Nothing
...
-          guideline_compat_for_all_deps) # 3
-
-    if !this_pr_can_use_special_jll_exceptions
-        push!(guidelines,
-              guideline_patch_release_does_not_narrow_julia_compat) # 4
-    end
+    checked_guidelines = Guideline[]
 
-    push!(guidelines,
-          guideline_allowed_jll_nonrecursive_dependencies) # 5
```
GunnarFarneback, line 72: Note: This guideline should only be applied to jll packages. It's a bug from #318 which is fixed in this PR. It's a bit worrying that the tests didn't catch it.

src/AutoMerge/new-version.jl
```diff
@@ -46,57 +47,49 @@ function pull_request_build(data::GitHubAutoMergeData, ::NewVersion)::Nothing
...
-    guidelines = Guideline[]
-    if !this_is_jll_package && pr_author_login ∉ data.authorized_authors
-        push!(guidelines,
-              guideline_jll_only_authorization) # 0
-    end
-
-    push!(guidelines,
-          guideline_pr_only_changes_allowed_files) # 1
+    jll_only_authorization = (!this_is_jll_package
+                              && pr_author_login ∉ data.authorized_authors)
```
ericphanson, line 51: I don't really understand the meaning of this variable; it seems negated from what I'd expect `jll_only_authorization` to mean. Or maybe I'm just not thinking clearly 😅. Mind explaining it a little?

src/AutoMerge/new-version.jl
ericphanson: good catch! looks like we did it right in `new-package.jl` but not here. As you mentioned, I think we are probably not covering that case in the integration tests. If you are up for it, you could look at adding one; the way it works is there's a list of folders that work like the "feature branch" a registration PR might make, and a list of "trunk branch" folders that correspond to possible states of the registry. Then we loop through and check that various pairs of feature-branch/trunk-branch pass or fail as appropriate.

src/AutoMerge/new-version.jl
GunnarFarneback: It means that the PR is for a non-jll package but the author only has authorization for jll packages. I'm open for ways to make it clearer, if someone can think of something better.

ericphanson: Ah ok. I think this one is more complicated because the check is being done in `applicable` instead of in the guideline. Maybe that could be addressed here?

DilumAluthge:
@ericphanson or @GunnarFarneback can you add an integration test that would catch the bug?

src/AutoMerge/new-version.jl
GunnarFarneback: Another option would be to exclude this authorization failure from the guideline checking and immediately return when it has been detected. That's done for other authorization failures.

DilumAluthge: That sounds fine. It would make the behavior consistent for all authorization failures.

DilumAluthge:
> @ericphanson or @GunnarFarneback can you add an integration test that would catch the bug?

BTW, if the integration test is the only thing holding this PR up, we can do it in a follow-up PR. I'd rather get this PR landed quickly so we can get 6.2.0 out the door, and we can add more integration tests at any time in the future.

src/AutoMerge/new-version.jl
DilumAluthge: > It means that the PR is for a non-jll package but the author only has authorization for jll packages. I'm open for ways to make it clearer, if someone can think of something better.

How about just add a comment above this line with exactly what you wrote here? I think that's sufficient.

ericphanson (approved):
lgtm

bors merge

ericphanson:
I'll pass this back to you @GunnarFarneback  😄. Since it's already approved, feel free to use `bors merge` so that it tests + merges if they pass.

bors delegate+

DilumAluthge:
Maybe you need `data.version` instead of `version`?

GunnarFarneback:
Yes.

bors merge

PR #342, opened 2021-01-17
Centralize authorization checks.

GunnarFarneback:
Do all authorization checks in one place.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>", "contributors"]
-version = "6.2.1"
+version = "6.3.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/src/AutoMerge/types.jl
+++ b/src/AutoMerge/types.jl
@@ -85,8 +85,8 @@ end
 # Constructor that requires all fields as named arguments.
 function GitHubAutoMergeData(;kwargs...)
     fields = fieldnames(GitHubAutoMergeData)
-    @assert Set(keys(kwargs)) == Set(fields)
-    @assert kwargs[:authorization] ∈ (:normal, :jll)
+    always_assert(Set(keys(kwargs)) == Set(fields))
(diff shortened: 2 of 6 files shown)


ericphanson (approved):

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>", "contributors"]
-version = "6.2.0"
+version = "6.3.0-DEV"
```
DilumAluthge, line 4: I generally prefer to release immediately after merging a PR. I.e. I only bump to a `-DEV` version if I'm planning on making multiple PRs in a row in a short period of time.

src/AutoMerge/types.jl
```diff
@@ -90,6 +86,7 @@ end
 function GitHubAutoMergeData(;kwargs...)
     fields = fieldnames(GitHubAutoMergeData)
     @assert Set(keys(kwargs)) == Set(fields)
+    @assert kwargs[:authorization] ∈ (:normal, :jll)
```
DilumAluthge, line 89: I notice there are a couple other places where we now use `@assert`.

In a future version of Julia, assertions will be disabled by default. Actually, there is currently a PR open to JuliaLang/julia that will disable assertions by default.

So, in this package, I think it's best if we don't use `@assert`. Instead, we have the `always_assert` function.

If we like, we could replace the `always_assert` function with an `@always_assert` macro, if that is easier to use.

DilumAluthge (changes requested):

src/AutoMerge/types.jl
ericphanson: Good point, I forgot about that

DilumAluthge:
bors merge

DilumAluthge:
@GunnarFarneback See the test failure above.

GunnarFarneback:
Yeah, things are not coming in quite the right order now.

GunnarFarneback:
The easiest way to solve it is to do the new package/version check before the authorization check, if everybody is fine with that.

DilumAluthge:
That's probably fine. We might eventually need to improve the `new_version_title_regex` and `new_package_title_regex` regexes, but we can do that in a different PR.

DilumAluthge:
BTW you'll need to resolve the merge conflict.

ericphanson:
bors merge

Issue #322, opened 2021-01-10
Re-enable the TagBotTriggers unit tests


DilumAluthge:
In https://github.com/JuliaRegistries/RegistryCI.jl/pull/320, I disabled the TagBotTriggers unit tests because they are failing on all Julia versions (1.3, 1.5, nightly).

We should figure out why they are failing, and re-enable them.

DilumAluthge:
cc: @christopher-dG

Issue #313, opened 2020-12-28
CI should require presence of test/runtest.jl for new packages


oxinabox:
A much weaker version of https://github.com/JuliaRegistries/RegistryCI.jl/issues/120
and similar to #261 

Even if we are not going to run it, or check that it contains something meaningful,
I think it is worth checking the bare minimum effort has been put into making sure the package is good.
At least be aware enough of what is going on to know that packages have these.

We used to have this requirement at one point for registring packages back in 0.4 days.

DilumAluthge:
Makes sense to me

KristofferC:
It seems unnecessary to add friction to registering a package that can just be circumvented with a `@test 1==1` (which is what people that didn't add any tests in the first place likely will do).

ericphanson:
The friction is only in the case where there are no tests, so if we want to encourage tests, that seems like a good place to add friction, no?

Sure, it can be easily circumvented, but it’s at least a nudge in the right direction. I think it would also help set the expectation that tests are needed, eg a new developer might not know they’re something they should have, and learn about it from an automerge failure.

In general I think it’s fine for automerge to be opinionated in order to try to help nudge the package ecosystem in a good direction (since RegistryCI’s biggest client, the General registry, is a shared resource we want to keep healthy, not just an easy way to access random pieces of code).

ericphanson:
With the technology of ~~~tokei~~~ (https://github.com/XAMPPRocky/tokei, packaged as `Tokei_jll`), we could count lines of Julia code in `/test` (ref #351 for `/src`). That could be used to create a more stringent requirement than just that `tests/runtests.jl` exists (but still pretty cirumventable, just write `@test true` a bunch of times).

My opinion is that we probably should **not** do this (at least at this point), and just start with checking that `test/runtests.jl` exists. IMO that covers most of the bases regarding getting new devs up to speed regarding the package expectations and where to put tests, without being too burdensome.

Issue #283, opened 2020-10-23
Add approved list of packages that skip installation test


odow:
Some packages, particularly in JuMP-dev, have commercial third-party dependencies that we can't provide binaries for. Examples include:
* https://github.com/jump-dev/CPLEX.jl
* https://github.com/jump-dev/Gurobi.jl
* https://github.com/jump-dev/Xpress.jl
* https://github.com/jump-dev/MosekTools.jl

This means such packages can't be `automerge`d because they fail the `Pkg.add` and `import` tests.

To prevent manual merging of every new release, the [proposal here](https://github.com/JuliaRegistries/General/pull/23397#issuecomment-714821893) is to have an approved list of packages somewhere that skip the `Pkg.add` and `import` tests:
https://github.com/JuliaRegistries/RegistryCI.jl/blob/88506ed3a6cae70d593b2eb05be79f97e7ddac63/src/AutoMerge/new-version.jl#L155-L165

Presumably this would be a file in `General` with the name of the package.

ericphanson:
I’m not sure how easy it would be to implement, but I think a good rule could be

- If it’s a new package, always require these tests
- if it’s a new version, and the previous registered version passes these tests, then require them
- if it’s a new version and the previous registered version fails these tests, then skip them 

Then we don’t need a list, and once you get manually merged once you can get auto merged, but if it’s a regression then it needs a manual merge.

DilumAluthge:
Just check if `get(ENV, "JULIA_REGISTRYCI_AUTOMERGE", "") == "true"`. If it is, don't try to load the commercial dependency.

DilumAluthge:
> * if it’s a new version and the previous registered version fails these tests, then skip them

I don't think this is a good idea. The goal of automerge is to be as conservative as possible. We should err on the side of being conservative and requiring manual merge.

DilumAluthge:
The best solution is to make your packages always loadable, even when the commercial dependency is not available.

For example, take a look here: https://github.com/JuliaGPU/CUDA.jl/blob/master/src/initialization.jl

CUDA.jl is always installable and loadable, even if you don't have a GPU.

That's the best solution. An alternative is to check if `get(ENV, "JULIA_REGISTRYCI_AUTOMERGE", "") == "true"`.

@fredrikekre Any thoughts?

odow:
> The best solution is to make your packages always loadable, even when the commercial dependency is not available.

[...]

PR #344, opened 2021-01-18
Add license check to automerge guidelines

### LINKED ISSUE #261: License checks
We should check the existence of a license file before merging new packages. We should also, ideally, check that the license is a valid open source licence before automerging. 

I noticed that recently a package was added without a license. https://github.com/JuliaRegistries/General/pull/18690

ericphanson:
fixes #261 by using [LicenseCheck.jl](https://github.com/ericphanson/LicenseCheck.jl) to wrap the Go library [licensecheck](https://github.com/google/licensecheck) in order to find the most likely license file and check what licenses it contains. LicenseCheck should be registered in a few hours; CI will fail until then.

License-file finding strategy: this is implemented in LicenseCheck.jl. The current strategy is: if there are more than 100 toplevel files, then check any that are in a long hardcoded list (COPYING, LICENSE, GPL.txt, etc). If there are fewer than 100, then check every plaintext file at toplevel. For each file we check, `licensecheck` returns a list of licenses it has found in that file (if any), and the percent of the file covered by those licenses it found (`percent_covered`). We take the file with the highest `percent_covered` and assume that is the license file of interest.

License rules: what I've currently implemented is: for each version being registered (not new packages only), there must be a file of which at least 70% of the contents are covered by identified licenses, and all of those licenses must be OSI-approved (determined by taking the `licensecheck`-obtained identifier, and seeing if it is' on a list of SDPX license identifiers which are OSI-approved, via LicenseCheck.jl).

Lots of choices in all of this, and maybe these rules will lead to too many falsely-flagged packages, I'm not sure.

The license guideline is enabled by a keyword argument to `run`, which is off by default so that this is not breaking. IMO we should just turn it on for General in it's config.

*update*: Now also looks for a `license` key in the Project.toml and takes that as the source of truth, if it exists.

---
original message

Addresses #261. It doesn’t necessarily close the issue, since this PR just does the first step, which is check if a license file exists at all. Checking the license against a list of OSI compatible licenses via licensee or other tools is left for future work.

The first commit moves a TagBot utility up to be used as a general RegistryCI utility, and the second uses it for a check if a license file exists.

This isn't done yet because it still needs a test for the license check; I'm not totally sure how to do that because it needs an actual PR body to parse and a repo to go look at, and doesn't just care about the commits to the registry itself. I suspect the TagBot tests that use mocking might be what I should look at, but they are currently broken (#322).

This is breaking because I set the license check to default to true, but we could also just set it to default to false and turn it on specifically for General, in which case it would be not breaking. (I did make sure it could be turned off since private registries may not use licenses for all their packages.)

By the way, adding new guidelines to the pipeline is super easy thanks to @GunnarFarneback's recent work!


ericphanson:
bors try

ericphanson:
bors try

ericphanson:
bors try

ericphanson:
Hmm, just reread https://github.com/JuliaRegistries/RegistryCI.jl/issues/261#issuecomment-720518808 and will need to change the approach here

src/AutoMerge/guidelines.jl
```diff
@@ -331,6 +331,40 @@ function meets_standard_initial_version_number(version)
...
+        return false, "Something went wrong: pull request body is `nothing`"
+    end
+
+    repo, version, commit = repo_version_commit_of_pull_request_body(repo; skip_if_JLL_in_body=false)
+    repo === nothing && return false, "Could not parse repository from PR body"
+    version === nothing && return false, "Could not parse version from PR body"
+    commit === nothing && return false, "Could not parse commit from PR body"
+
+    files, pages = try
+        GitHub.directory(repo, ""; auth=data.auth, ref=commit)
```
fredrikekre, line 347: IIRC the API has a license endpoint. Perhaps use that and let GitHub find the file instead of trying the filenames below?

ericphanson: Yeah, that was suggested in https://github.com/JuliaRegistries/RegistryCI.jl/issues/261#issuecomment-720518808 too. I think I'd prefer to use @DilumAluthge's other suggestion there and use `Pkg.add` to get the code instead of the GitHub API, to allow non-github packages.

I don't really see the advantage to using GitHub's license detection for github packages and a hardcoded list for non-github packages; is that it that GitHub packages are the common case and the license detection will have better chances at finding the license?

ericphanson: I now use the code from Pkg.add, and use `licensecheck` to identify the license (discussed more in the OP)

ericphanson:
@oxinabox suggested using a `license` field in the package's Project.toml, if it exists, to override the license detection here. That field is mentioned in passing in https://julialang.github.io/Pkg.jl/dev/glossary/ but not in https://julialang.github.io/Pkg.jl/dev/toml-files/#Project.toml.

ericphanson:
the latest commit adds a check to see if there is a `license` key in the Project.toml, with either the SDPX 3.10 identifier of an OSI-approved license, or a list of such identifiers (in which case all must be OSI-approved for the check to pass, as currently written). This list can be seen here: https://github.com/ericphanson/LicenseCheck.jl/blob/main/src/OSI_LICENSES.jl. It includes things like "MIT", "GPL-2.0" etc, so it should be fairly intuitive to use.

If the Project.toml has a license key, we do not try to find a license file ourselves; i.e. we take the Project as a source of truth. This can help if the autodetection goes wrong; the solution is to just put a license identifier in the Project.toml, and then there's no way to confuse it. It's helpful to have such an opt-out because the check is done on every version, so the same package could get repeatedly flagged by accident otherwise.

That being said, I do think the autodetection should be pretty good here.

ericphanson:
bors try

(still needs tests for the new functionality, but let's see if I broke the old functionality...)

ericphanson:
bors try

ericphanson:
The license check requires access to the package code in order to look and see if there is a license there (and to parse the Project.toml to look for a license field). This functionality (looking at the package code) would also be needed for #313. 

My previous approach (i.e. in the 2nd to last commit) was to add in a `process_pkgdir` callback to `_run_pkg_commands` that would run after the commands were run, but before the tempdirs were cleaned up. This would provide an opportunity to inspect the downloaded code. However, this approach meant that either every check that wants access to the pkg code needs to redownload it (which increases the chances of hitting a network failure etc), or all the checks that want access to the pkg code would need to be run within the same "guideline".

In the latest commit, I instead added a `Ref` field to `GitHubAutoMergeData` that would allow the `Pkg.add` step to populate the directory with the tempdir, and pass a cleanup callback to be called after the checks were done. That way, all checks that want access to the package code can re-use it (by accessing this field), while keeping the checks somewhat atomic.

This makes `GitHubAutoMergeData` stateful, which is not ideal IMO. However, at least it's somewhat local state being passed around explicitly.

@GunnarFarneback, since you designed the new guidelines loop, would you mind taking a look and letting me know what you think?

bors try

ericphanson:
bors try

GunnarFarneback:
I would do it like this:
1. Add a field for the directory in GitHubAutoMergeData and initialize it up front with `mktempdir`.
2. Download the package code when needed.
3. Don't worry about cleanup. From Julia 1.3, directories created by `mktempdir` clean up themselves on process exit.

ericphanson:
Ah ok, that sounds good. So when we download the code, we'll use that field to get the tempdir to download it to, and when we use it, we just use that directory too.

ericphanson:
bors try

ericphanson:
bors try

ericphanson:
bors try

ericphanson:
Ok, I think this just needs integration tests and it's good to go from my end. Will have to think a bit about how to do that...

DilumAluthge:
I'm not a fan of the `license` key in the Project.toml file. It's not legally binding, right? Someone could add `license = MIT` in their Project.toml file, but not have a LICENSE file in their repo, and then later claim in court that their software did not have an open-source license.

DilumAluthge:
In contrast, an actual license file containing the text of an actual license - this is legally binding.

So we should only use that.

ericphanson:
I mentioned this concern to @oxinabox on Slack, who said

> Intellectual property is governed by civil law.
> Civil law actually let's you get away with very little in practice, because the judges are not bound letter of the law and take into > account sensibility, and intent.
> If someone registers a package in the public registry and sets the license field it's hard to say they were not intending for > people to think that they can use it under that lisence.
> (IANAL)

(hope you don't mind me quoting you here Lyndon, I can delete if you object). I thought that made sense, and that having such a field will give us an easy fix if there are false positives; instead of needing a manual merge every new version, we can say just add the license field with the actual license.

That being said, the license checking code should be pretty robust; I literally check every top-level plain text file for any licenses if there are fewer than 100 toplevel files, and if there are more than that then there's a list of 500 names I check for. I think where there could be false positives is if

* the license is not at top-level in the package, e.g. if the package is in a subdirectory and the license is top-level in the repo
* if the license is embedded in a larger text file; the check currently requires that 70% of the file the license contains is consists of recogonized licenses. Perhaps there are situations where there's a legitimate reason to have other text in the same file.

FWIW, AnalyzeRegistry currently reports 260 (out of ~4.4k) non-jll packages without a recognized license (checked the same way as this PR does), so either there's a lot of packages without a license or there's a lot of false positives.

oxinabox:
I am in stong agreement with my past self, though I am still not a lawyer.
But if you try and sue someone after you have intentionally mislead them into breaking contract with you.
You are going to get laughed out of court. Things like the clean hands doctorine apply:

>"to derive advantage from his dishonest conduct in so direct a manner that it is considered to be unjust to grant him relief".

More broadly, IP law is civil law, the whole notion of legally binding is no where near as clear cut as people make it out to be. It goes to a court which considers precident and intent, and deliberates.
A court is not an algorithm exectuting the contract like it were a test suite.

but before we even talk about that, we should also appreciate that this a CI check to catch easy mistakes; it's it self is not presenting a legal opinion at all.
The registry doesn't promise its users that all registered packages are packages they are allowed to use.
It doen't have to be perfect and the check isn't going to be perfect.

Theoretical example of where it will fail:
Like what if the LISCENSE.md said GPL, but when you started the program it printed to the screen "Actually, this is also MIT", and the readme and docs said the same.
We are not going to catch that.
But we are going to catch 99% of cases with the check on LISCENSE.md.
and the lisence Project.toml  serves as a good override.
It seems like the sensible place, after all it is where all the other things we want to tell the registry are.
we could have another file if wanted `Nonlegallybinding_registry_overrides.toml`

ericphanson:
> but before we even talk about that, we should also appreciate that this a CI check to catch easy mistakes; it's it self is not presenting a legal opinion at all.

I think this is a good point; IMO a light touch is good to encourage best-practices without being burdensome. Since we can't guarantee anything and it's not a "curated" registry, I think it's good to have escape hatches like this license field (or the workaround for #283 where packages can skip loading a binary dependency if it's not packaged so they are importable).

DilumAluthge:
How about this as a compromise:

Instead of the value of the `license` key being the name of the license, how about it is the relative path to the LICENSE file. Then, we read in that file and parse it.

So e.g. some examples might be:

```toml
license = "LICENSE.md"
``` 

```toml
license = "foo/bar/LICENSE.md"
``` 

We determine the absolute path to the license via the following:
```julia
path_to_project_file = ...
parsed_project_file = TOML.parsefile(path_to_project_file)
relative_path_to_license_file = parsed_project_file["license"]
@assert !occursin("..", relative_path_to_license_file)
absolute_path_to_license_file = joinpath(dirname(path_to_project_file), relative_path_to_license_file)
```

Then we open `absolute_path_to_license_file` and parse it using your license parsing code.

This gives us the best of both worlds. Users can provide the `license` key as an override in those false-positive cases you describe. But it still requires that they have the full text of the license in their package.

ericphanson:
On a technical level, I don't that that addresses the case of "the license is in the top-level of a repo, and the package is in a subdirectory", since Pkg.add only gives us the code for the package itself, not the whole repo. We could switch to cloning the repo instead of Pkg.add, but that adds complexity

DilumAluthge:
I'm not comfortable with the `license = "MIT"` approach. I think that if you want AutoMerge, you should be willing to have the full text of the license inside your package.

DilumAluthge:
> On a technical level, I don't that that addresses the case of "the license is in the top-level of a repo, and the package is in a subdirectory", since Pkg.add only gives us the code for the package itself, not the whole repo. We could switch to cloning the repo instead of Pkg.add, but that adds complexity

Correct. In this case, if the user wants to have AutoMerge, they also need to put a copy of the license in the same folder as their `Project.toml` file. This is easy to do, and they only need to do it once.

ericphanson:
> I'm not comfortable with the license = "MIT" approach. I think that if you want AutoMerge, you should be willing to have the full text of the license inside your package.

Does it help that it's the SDPX identifier of the license, not an informal name? I.e. there's a precise meaning to what `MIT` means here.

DilumAluthge:
I really want the full text of the OSI-approved license to be somewhere in the package source code.

oxinabox:
Its convoluted and confusing.

IANAL but as i understand it there isn't actually even a need to include the full text of a liscense in your repo in the first place, from a IP law perspective.
That is just a term that some lisences include, (in which case derivatives do need to).
It probably has a historical cause of pre-internet you couldn't just go and look these things up online.
The thing that is required from a liscense is to communciate clearly what your terms are.
A lot of stuff under creative commons liscenses just use a little graphic that is a hyperlink.

ericphanson:
Do we know why the `license` field was removed from the Project.toml format? @oxinabox mentioned it used to be there and is mentioned in the glossary but not in the list of keys in the Pkg docs. Maybe knowing the reasoning for that that can help inform us the discussion here.

DilumAluthge:
What if someone says `license = "BSD"`? Which BSD license do they mean?

Including the full text of the license is unambiguous as to the terms of the software.

ericphanson:
> What if someone says `license = "BSD"`? Which BSD license do they mean?
> 
> Including the full text of the license is unambiguous as to the terms of the software.

```julia
julia> using LicenseCheck

julia> is_osi_approved("BSD")
false
```
not allowed 🙂

DilumAluthge:
I really don't understand why it's so inconvenient to stick a copy of the MIT License (or whatever license you want) in the top level of your package, and then point the `license` key to that file.

You only need to do it once per package.

We require upper-bounded compat entries for AutoMerge. That's much more inconvenient, since you need to constantly be bumping your compat entries. But we require it for AutoMerge.

DilumAluthge:
GitHub says (https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/licensing-a-repository#determining-the-location-of-your-license)>
> As a best practice, we encourage you to include the license file with your project.

One of the main goals of AutoMerge is to enforce best practices across the ecosystem. I'm comfortable saying that this is a best practice that I want to require, at least if you want AutoMerge.

ericphanson:
Yeah, that sounds fair enough to me 🤷. Would be nice to see if we can come to a consensus though, I think the discussion is still helpful. E.g. what is the use-case you're thinking of @oxinabox that would motivate the field? false-postives or something else?

ericphanson:
I think though if we want to support the path version, we should call it `license_path` instead of `license`, espescially since apparently historically there used to be an official `license` field where you put the name/identifier.

DilumAluthge:
> I think though if we want to support the path version, we should call it `license_path` instead of `license`, espescially since apparently historically there used to be an official `license` field where you put the name/identifier.

Definitely.

oxinabox:
NPM uses exactly the approach suggested here, without looking at the LICENSE.md
It just looks for SPDX license identifier in the package.json.
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license

Noting that NPM has over 1 million packages,
and that thus most likely *has* had someone legal work it over (probably dozens of times).


If a SPDX identifier in a field,  is good enough for NPM then I think it is good enough for us to use as a fallback/override.

DilumAluthge:
Hmmm. Personally I still would like to enforce that the license text is in the repo.

But the NPM example is some good prior art.

Given what you've found for NPM, let's see what some other people think, e.g. @fredrikekre and @KristofferC. If other people are fine with the system, I'll defer to the majority.

DilumAluthge:
Also pinging @aviks to see what he thinks.

fredrikekre:
IMO it is fine, but if we go that route it should be documented in Pkg as an official field in the project file. Perhaps it would be good to check that the field-specified license and the LICENSE-specified license agree (if they both exist).

Shouldn't this check be executed for every PR btw, not just for new packages?

ericphanson:
> Shouldn't this check be executed for every PR btw, not just for new packages?

Yeah, it is for every PR

fredrikekre:
Ah ok. What are examples of false positives? What happens with e.g. a file like this: https://github.com/fredrikekre/Literate.jl/blob/master/LICENSE.md

ericphanson:
> Ah ok. What are examples of false positives? What happens with e.g. a file like this: https://github.com/fredrikekre/Literate.jl/blob/master/LICENSE.md

That passes without an issue, e.g. using AnalyzeRegistry which uses the same check,
```julia
julia> using AnalyzeRegistry

julia> result = analyze_from_registry(find_packages("Literate")[1])
Package Literate:
  * repo: https://github.com/fredrikekre/Literate.jl.git
  * uuid: 98b081ad-f1c9-55d3-8b20-4c87d4299306
  * is reachable: true
  * lines of Julia code in `src`: 727
  * lines of Julia code in `test`: 1047
  * has license(s) in file: MIT, MIT, MIT
    * filename: LICENSE.md
    * OSI approved: true
  * has documentation: true
  * has tests: true
  * has continuous integration: true
    * GitHub Actions

julia> result.license_files[1]
(license_filename = "LICENSE.md", licenses_found = ["MIT", "MIT", "MIT"], license_file_percent_covered = 88.62068965517241)
```
Here, all OSI approved licenses are found, and the file is more than 70% covered by them, so no problem.

DilumAluthge:
Can you show us an example of something that doesn't pass?

oxinabox:
>  E.g. what is the use-case you're thinking of @oxinabox that would motivate the field? false-positives or something else?


We mostly need to worry about things getting confused by long liscenses.
Primarily with things being rected because there LISCENSE.md is complicated.
Try julia's maybe?
https://github.com/JuliaLang/julia/blob/master/LICENSE.md

Interestingly NPM advises closed source stuff to set:
```
"license": "UNLICENSED"
```
which i guess could be a useful guard if you are in the process of open sourcing something, but don't want this commit to be open sourced.
Or if you are doing some non-open source, but source public thing like MongoDB does.

ericphanson:
> Can you show us an example of something that doesn't pass?

I can do you one better: here are all 351 packages in the General registry that would not pass this check:

<details>

* https://github.com/MichaelLLi/MatrixImpute.jl.git
* https://github.com/JuliaClimate/ClimateTools.jl.git
* https://github.com/cossio/SimpleDataLoader.jl.git
* https://github.com/cossio/OneHot.jl.git
* https://github.com/kindlychung/SuperEnum.jl.git
* https://github.com/briochemc/GEOTRACES.jl.git
* https://github.com/cossio/LocalStore.jl.git
* https://github.com/RoboticExplorationLab/RobotZoo.jl.git
* https://github.com/tom--lee/MolecularTopologies.jl.git
* https://github.com/alan-turing-institute/MLJFlux.jl.git
* https://github.com/mossr/AddPackage.jl.git
* https://github.com/ajwheeler/TheCannon.jl.git
* https://github.com/abhinavmehndiratta/LightGraphsGraphBLAS.jl.git
* https://github.com/FluxML/Torch.jl.git
* https://github.com/portugueslab/DeepDish.jl.git
* https://github.com/wookay/EmojiSymbols.jl.git
* https://github.com/JuliaGraphics/ColorSchemes.jl.git
* https://github.com/pevnak/Duff.jl.git
* https://github.com/Yujie-W/ConstrainedRootSolvers.jl.git
* https://github.com/SciML/DASKR.jl.git
* https://github.com/PumasAI/ProvenanceBase.jl.git
* https://github.com/MUsmanZahid/TeXTable.jl.git
* https://github.com/zemjulia/FEHM.jl.git
* https://github.com/mcabbott/NamedPlus.jl.git
* https://github.com/jump-dev/SeDuMi.jl.git
* https://github.com/mossr/Tetris.jl.git
* https://github.com/dss-extensions/OpenDSSDirect.jl.git
* https://github.com/pauldubois98/ModularFormsModuloTwo.jl.git
* https://github.com/TuringLang/MCMCDebugging.jl.git
* https://github.com/PSORLab/DynamicBoundsBase.jl.git
* https://github.com/maartenvd/MPSKitModels.jl.git
* https://github.com/PetrKryslUCSD/MeshPorter.jl.git
* https://github.com/loooj58/PiecewisePolynomials.jl.git
* https://github.com/baggepinnen/TotalLeastSquares.jl.git
* https://github.com/Yujie-W/PlotPlants.jl.git
* https://github.com/mipals/SymSemiseparableMatrices.jl.git
* https://github.com/jakobnissen/Probably.jl.git
* https://github.com/jump-dev/MINLPTests.jl.git
* https://github.com/Yujie-W/SoilPlantAirContinuum.jl.git
* https://github.com/owinter92/SimpleTweaks.jl.git
* https://github.com/mewertd2/Handles.jl.git
* https://github.com/jump-dev/SDPNAL.jl.git
* https://github.com/the-sushi/MixFit.jl.git
* https://github.com/Yujie-W/WaterPhysics.jl.git
* https://github.com/dmbates/NLreg.jl.git
* https://github.com/JockLawrie/Authorization.jl.git
* https://github.com/JuliaDynamics/CausalityToolsBase.jl.git
* https://github.com/tmthyln/CEDICT.jl.git
* https://github.com/rjdverbeek-tud/Atmosphere.jl.git
* https://github.com/PainterQubits/UnitfulUS.jl.git
* https://github.com/tom--lee/Cookbook.jl.git
* https://github.com/ollin18/Node2Vec.jl.git
* https://github.com/usnistgov/NeXLCore.jl.git
* https://github.com/JuliaIntervals/SetRounding.jl.git
* https://github.com/digitaldomain/DiscreteDifferentialGeometry.jl.git
* https://github.com/gridap/GridapPardiso.jl.git
* https://github.com/jump-dev/ComplexOptInterface.jl.git
* https://github.com/johncwok/SpectralEnvelope.jl.git
* https://github.com/anders-dc/Granular.jl.git
* https://github.com/oscar-system/LoadFlint.jl.git
* https://github.com/bat/UltraNest.jl.git
* https://github.com/runtosolve/StructuresKit.jl.git
* https://github.com/JuliaGeometry/Triangulate.jl.git
* https://github.com/AStupidBear/GCMAES.jl.git
* https://github.com/tom--lee/MolecularTrajectories.jl.git
* https://github.com/CarloLucibello/Erdos.jl.git
* https://github.com/StanJulia/StanMCMCChain.jl.git
* https://github.com/ChrisRackauckas/VectorizedRoutines.jl.git
* https://github.com/EchoJulia/EchogramColorSchemes.jl.git
* https://github.com/FermiQC/TBLIS.jl.git
* https://github.com/sglyon/ORCA.jl.git
* https://github.com/JuliaGPU/oneAPI.jl.git
* https://github.com/slmcbane/MirroredArrayViews.jl.git
* https://github.com/simon-lc/AlgamesDriving.jl.git
* https://github.com/pevnak/Unitary.jl.git
* https://github.com/mrtkp9993/Bioinformatics.jl.git
* https://github.com/KlausC/ChannelBuffers.jl.git
* https://github.com/kellertuer/OpenPixelControl.jl.git
* https://github.com/Quantum-Many-Body/ExactDiagonalization.jl.git
* https://github.com/prassee/DactParser.jl.git
* https://github.com/lanl-ansi/PowerWaterModels.jl.git
* https://github.com/Nauss/Repotomata.git
* https://github.com/IanButterworth/SystemBenchmark.jl.git
* https://github.com/JockLawrie/GoogleCloudObjectStores.jl.git
* https://github.com/RoboticExplorationLab/TrajOptPlots.jl.git
* https://github.com/AStupidBear/PandasLite.jl.git
* https://github.com/JuliaReinforcementLearning/ReinforcementLearningEnvironments.jl.git
* https://github.com/lpfann/Fri.jl.git
* https://github.com/lanl-ansi/PowerModelsRestoration.jl.git
* https://github.com/blegat/SetProg.jl.git
* https://github.com/pfarndt/EnergyStatistics.jl.git
* https://github.com/DhairyaLGandhi/UNet.jl.git
* https://github.com/jamesrhester/CrystalInfoFramework.jl.git
* https://github.com/JeffreySarnoff/SingleFloats.jl.git
* https://github.com/wookay/Octo.jl.git
* https://github.com/ronisbr/TextUserInterfaces.jl.git
* https://github.com/haampie/StructuredGrids.jl.git
* https://github.com/lanl-ansi/PetroleumModels.jl.git
* https://github.com/johncwok/IntegerIB.jl.git
* https://github.com/gridap/MiniQhull.jl.git
* https://github.com/cjdoris/LTWA.jl.git
* https://github.com/usnistgov/NeXLMatrixCorrection.jl.git
* https://github.com/JuliaHomotopyContinuation/ProjectiveVectors.jl.git
* https://github.com/rahumble/ClusterRunner.jl.git
* https://github.com/j-fu/VTKView.jl.git
* https://github.com/nosewings/BigMacro.jl.git
* https://github.com/JonathanCourtois/MackeyGlass.jl.git
* https://github.com/JuliaMatrices/BandedMatrices.jl.git
* https://github.com/cossio/ROCCurves.jl.git
* https://github.com/ceandrade/BrkgaMpIpr.jl.git
* https://github.com/mewertd2/Blades.jl.git
* https://github.com/JuliaReinforcementLearning/Hanabi.jl.git
* https://github.com/mcabbott/Bhaskara.jl.git
* https://github.com/longemen3000/CachedFunctions.jl.git
* https://github.com/Tokazama/TimeAxes.jl.git
* https://github.com/byuflowlab/VortexLattice.jl.git
* https://github.com/briochemc/OceanographyCruises.jl.git
* https://github.com/GuilhermeHaetinger/KelvinletsImage.jl.git
* https://github.com/PetrKryslUCSD/FinEtoolsFlexBeams.jl.git
* https://github.com/cossio/SequenceLogos.jl.git
* https://github.com/felipenoris/LDAPClient.jl.git
* https://github.com/floswald/SMM.jl.git
* https://github.com/kraftpunk97/GymSpaces.jl.git
* https://github.com/johncwok/ChangePointDetection.jl.git
* https://github.com/bcbi/CountdownNumbers.jl.git
* https://github.com/JuliaReinforcementLearning/ReinforcementLearningBase.jl.git
* https://github.com/johncwok/DCCA.jl.git
* https://github.com/femtomc/Problox.jl.git
* https://github.com/RoboticExplorationLab/Altro.jl.git
* https://github.com/maartenvd/SUNRepresentations.jl.git
* https://github.com/AStupidBear/Backtester.jl.git
* https://github.com/dfdx/Avalon.jl.git
* https://github.com/MichaelHatherly/CommonMark.jl.git
* https://github.com/JuliaGizmos/WebToys.jl.git
* https://github.com/adrhill/NoiseRobustDifferentiation.jl.git
* https://github.com/hardbyte/Paillier.jl.git
* https://github.com/usnistgov/FFAST.jl.git
* https://github.com/PetrKryslUCSD/MeshKeeper.jl.git
* https://github.com/NTimmons/ApproximationAnalysis.jl.git
* https://github.com/hros/XXhash.jl.git
* https://github.com/onetonfoot/ExprManipulation.jl.git
* https://github.com/Quantum-Many-Body/QuantumLattices.jl.git
* https://github.com/kailaix/ADCMEKit.jl.git
* https://github.com/TensorDecompositions/PhysicsInformedML.jl.git
* https://github.com/jump-dev/SDPT3.jl.git
* https://github.com/mlpack/mlpack.jl.git
* https://github.com/lanl-ansi/GraphicalModelLearning.jl.git
* https://github.com/goretkin/FixArgs.jl.git
* https://github.com/JuliaString/PCRE2.jl.git
* https://github.com/FHell/EmbeddedGraphs.jl.git
* https://github.com/carlobaldassi/ArgParse.jl.git
* https://github.com/usnistgov/NeXLUncertainties.jl.git
* https://github.com/eduardsmetanin/NoisyQuantumComputerSimulator.git
* https://github.com/JuliaString/StrRegex.jl.git
* https://github.com/bjack205/SolverLogging.jl.git
* https://github.com/AStupidBear/FluxUtils.jl.git
* https://github.com/lanl-ansi/ThreePhasePowerModels.jl.git
* https://github.com/JockLawrie/SpineBasedRecordLinkage.jl.git
* https://github.com/baggepinnen/AutomaticDocstrings.jl.git
* https://github.com/NREL/REoptLite.git
* https://github.com/EcoJulia/EcologicalNetworksPlots.jl.git
* https://github.com/CliMA/Cloudy.jl.git
* https://github.com/Yujie-W/StomataModels.jl.git
* https://github.com/PainterQubits/UnitfulIntegration.jl.git
* https://github.com/abhinavmehndiratta/GraphBLASInterface.jl.git
* https://github.com/FermiQC/Lints.git
* https://github.com/JuliaLang/julia.git
* https://github.com/EliasBcd/RevealedPreferences.jl.git
* https://github.com/JaneliaSciComp/NIDAQ.jl.git
* https://github.com/cossio/Pfam.jl.git
* https://github.com/JuliaString/Strs.jl.git
* https://github.com/EmilioAlvizo/TermoQuimica.git
* https://github.com/markushhh/YahooFinance.jl.git
* https://github.com/madsjulia/AffineInvariantMCMC.jl.git
* https://github.com/ScottishCovidResponse/DataRegistryUtils.jl.git
* https://github.com/JuliaNeighbors/Neighborhood.jl.git
* https://github.com/louiscmm/Sinaica.jl.git
* https://github.com/PetrKryslUCSD/MeshMaker.jl.git
* https://github.com/rdeits/EdgeCameras.jl.git
* https://github.com/KlausC/TarIterators.jl.git
* https://github.com/abhinavmehndiratta/SuiteSparseGraphBLAS.jl.git
* https://github.com/JuliaImages/ImageShow.jl.git
* https://github.com/xukai92/Catsay.jl.git
* https://github.com/Tokazama/FieldProperties.jl.git
* https://github.com/sambitdash/PDFIO.jl.git
* https://github.com/johncwok/SerialDependence.jl.git
* https://github.com/femtomc/Pyrox.jl.git
* https://github.com/JoshuaPurtell/SocialSolver.jl.git
* https://github.com/bcbi/CountdownLetters.jl.git
* https://github.com/jaypmorgan/Adversarial.jl.git
* https://github.com/biaslab/ForneyLab.jl.git
* https://github.com/JuliaGraphics/NamedColors.jl.git
* https://github.com/lanl-ansi/GasModels.jl.git
* https://github.com/PtFEM/PtFEM.jl.git
* https://github.com/NormannR/LRESolve.jl.git
* https://github.com/phelipe/RobotDescriptions.jl.git
* https://github.com/femtomc/GenFlux.jl.git
* https://github.com/JuliaGraphs/LightGraphs.jl.git
* https://github.com/RoboticExplorationLab/RobotDynamics.jl.git
* https://github.com/rapus95/HubCLI.jl.git
* https://github.com/lanl-ansi/PowerModelsSecurityConstrained.jl.git
* https://github.com/lanl-ansi/GasPowerModels.jl.git
* https://github.com/lanl-ansi/InfrastructureModels.jl.git
* https://github.com/baggepinnen/ThreadTools.jl.git
* https://github.com/sp94/LorentzDrudeMetals.jl.git
* https://github.com/JockLawrie/ObjectStores.jl.git
* https://github.com/ShozenD/AutocorrelationShell.jl.git
* https://github.com/j-fu/GridVisualize.jl.git
* https://github.com/simon-lc/Algames.jl.git
* https://github.com/JuliaDocs/Highlights.jl.git
* https://github.com/mcabbott/TransmuteDims.jl.git
* https://github.com/TuringLang/MCMCChain.jl.git
* https://github.com/jamesrhester/DrelTools.jl.git
* https://github.com/JuliaGNSS/TrackingLoopFilters.jl.git
* https://github.com/jwscook/SeriesAccelerators.jl.git
* https://github.com/rimbach/Ccluster.jl.git
* https://github.com/zenna/Callbacks.jl.git
* https://github.com/cjdoris/ARFFFiles.jl.git
* https://github.com/oxfordcontrol/CDCS.jl.git
* https://github.com/haampie/IncompleteLU.jl.git
* https://github.com/Yujie-W/PkgUtility.jl.git
* https://github.com/s-valent/dichotomy.git
* https://github.com/ExpandingMan/LSPNeovim.jl.git
* https://github.com/JuliaDSP/DSP.jl.git
* https://github.com/tlienart/MLJScikitLearn.jl.git
* https://github.com/digitaldomain/Multivectors.jl.git
* https://github.com/PSORLab/DynamicBoundspODEsIneq.jl.git
* https://github.com/JuliaActors/ActorInterfaces.jl.git
* https://github.com/tom--lee/FormattedTables.jl.git
* https://github.com/madsjulia/FDDerivatives.jl.git
* https://github.com/bmourrain/MultivariateSeries.jl.git
* https://github.com/StatisticalRethinkingJulia/StatisticalRethinking.jl.git
* https://github.com/NTimmons/XboxController.jl.git
* https://github.com/CliMA/GriddingMachine.jl.git
* https://github.com/dcarrera/Planets.jl.git
* https://github.com/hzgzh/LiBr.jl.git
* https://github.com/aramirezreyes/AvailablePotentialEnergyFramework.jl.git
* https://github.com/lanl-ansi/MomentOpt.jl.git
* https://github.com/arzwa/NewickTree.jl.git
* https://github.com/NHDaly/ApplicationBuilderAppUtils.jl.git
* https://github.com/JuliaDynamics/CrossMappings.jl.git
* https://github.com/marekdedic/ThreadedIterables.jl.git
* https://github.com/lanl-ansi/PowerModelsDistribution.jl.git
* https://github.com/Nemocas/AbstractAlgebra.jl.git
* https://github.com/Yujie-W/CanopyLayers.jl.git
* https://github.com/TuringLang/MCMCChains.jl.git
* https://github.com/chengchingwen/Pickle.jl.git
* https://github.com/Jutho/OptimKit.jl.git
* https://github.com/luap-pik/SyntheticNetworks.git
* https://github.com/baggepinnen/DiskDataProviders.jl.git
* https://github.com/DrKrar/GeoEfficiency.jl.git
* https://github.com/lucianolorenti/StructDatabaseMapping.jl.git
* https://github.com/mossr/Snake.jl.git
* https://github.com/KlausC/PkgVersion.jl.git
* https://github.com/femtomc/SPPL.jl.git
* https://github.com/digitaldomain/HalfEdges.jl.git
* https://github.com/invenia/Models.jl.git
* https://github.com/JuliaMath/DecFP.jl.git
* https://github.com/mschauer/PointProcessInference.jl.git
* https://github.com/akaysh/SpiderMonkey.jl.git
* https://github.com/NLaws/LinDistFlow.git
* https://github.com/sisl/SOFA.jl.git
* https://github.com/wookay/Poptart.jl.git
* https://github.com/Yujie-W/PlantHydraulics.jl.git
* https://github.com/wangjie212/ChordalGraph.git
* https://github.com/JuliaNLSolvers/LineSearches.jl.git
* https://github.com/mlubin/JuMPChance.jl.git
* https://github.com/fireofearth/ModalIntervalArithmetic.jl.git
* https://github.com/s-valent/gradient-methods.git
* https://github.com/cjdoris/Edlib.jl.git
* https://github.com/Nemocas/Nemo.jl.git
* https://github.com/JuliaImages/ImageNoise.jl.git
* https://github.com/NTimmons/FastActivations.jl.git
* https://github.com/JuliaApproximation/OscillatoryIntegrals.jl.git
* https://github.com/JockLawrie/LocalDiskObjectStores.jl.git
* https://github.com/oscardssmith/BetterExp.jl.git
* https://github.com/MUsmanZahid/DTALib.jl.git
* https://github.com/mewertd2/KVectors.jl.git
* https://github.com/ayoublasri/Biomodelling.jl.git
* https://github.com/dmlc/MXNet.jl.git
* https://github.com/ranjanan/MonteCarloIntegration.jl.git
* https://github.com/lanl-ansi/WaterModels.jl.git
* https://github.com/jamesrhester/Lerche.jl.git
* https://github.com/felipenoris/BLPData.jl.git
* https://github.com/PainterQubits/Unitful.jl.git
* https://github.com/dharasim/TikzQTrees.jl.git
* https://github.com/OpenModelica/DoubleEnded.jl.git
* https://github.com/jstrube/LCIO.jl.git
* https://github.com/rob-luke/NIRX.jl.git
* https://github.com/lanl-ansi/Alpine.jl.git
* https://github.com/PetrKryslUCSD/MeshFinder.jl.git
* https://github.com/dellison/LargeMovieReviewDataset.jl.git
* https://github.com/s-valent/ConjugateGradientMethod.git
* https://github.com/zenna/Spec.jl.git
* https://github.com/malmaud/TensorFlow.jl.git
* https://github.com/Moelf/BigG.jl.git
* https://github.com/hakkelt/FunctionOperators.jl.git
* https://github.com/felipenoris/JSONWebTokens.jl.git
* https://github.com/RalphAS/GenericSchur.jl.git
* https://github.com/AStupidBear/StandardMarketData.jl.git
* https://github.com/etorkia/MCHammer.jl.git
* https://github.com/JuliaPlots/PlotUtils.jl.git
* https://github.com/briochemc/OceanBasins.jl.git
* https://github.com/OpenModelica/ImmutableList.jl.git
* https://github.com/kyungminlee/SuperLU.jl.git
* https://github.com/rbalexan/InfrastructureSensing.jl.git
* https://github.com/mattg3004/Schistoxpkg.jl.git
* https://github.com/gridap/GridapPETSc.jl.git
* https://github.com/dfdx/Lilith.jl.git
* https://github.com/JuliaDiff/HyperDualNumbers.jl.git
* https://github.com/usnistgov/NeXLSpectrum.jl.git
* https://github.com/lanl-ansi/PowerModels.jl.git
* https://github.com/JuliaAlgebra/MultivariateBases.jl.git
* https://github.com/gridap/GridapEmbedded.jl.git
* https://github.com/lucianolorenti/Estapir.jl.git
* https://github.com/cossio/ZygoteStructArrays.jl.git
* https://github.com/lanl-ansi/PowerModelsAnnex.jl.git
* https://github.com/rjsummers/Microwaves.jl.git
* https://github.com/OpenModelica/OMJulia.jl.git
* https://github.com/ChrisRackauckas/ParallelDataTransfer.jl.git
* https://github.com/oscar-system/GAPTypes.jl.git
* https://github.com/phaverty/GenomicVectors.jl.git
* https://github.com/kalmarek/PermutationGroups.jl.git
* https://github.com/cossio/BioSeqInt.jl.git
* https://github.com/sadit/KCenters.jl.git
* https://github.com/RBMpro/RestrictedBoltzmannMachines.jl.git
* https://github.com/JuliaParallel/UCX.jl.git
* https://github.com/alvsierra286/OOESAlgorithm.jl.git
* https://github.com/Tokazama/AxisTables.jl.git
* https://github.com/stakaz/BracedErrors.jl.git
* https://github.com/JaneliaSciComp/Morton.jl.git
* https://github.com/JuliaLang/MbedTLS.jl.git
* https://github.com/madsjulia/Wells.jl.git
* https://github.com/NTimmons/NormalMaps.jl.git
* https://github.com/dronir/QuaternionIntegrator.jl.git
* https://github.com/tehrengruber/MemberFunctions.jl.git
* https://github.com/EconForge/Dolang.jl.git
* https://github.com/akdor1154/NamedPositionals.jl.git
* https://github.com/ajwheeler/FITSTables.jl.git
* https://github.com/kul-forbes/CIAOAlgorithms.jl.git
* https://github.com/r0cketr1kky/Dogsay.jl.git
* https://github.com/lanl-ansi/PowerModelsAnalytics.jl.git
* https://github.com/neelsmith/PolytonicGreek.jl.git
* https://github.com/andreasnoack/FastPolynomialRoots.jl.git
* https://github.com/usnistgov/BoteSalvatICX.jl.git
* https://github.com/onetonfoot/Plumber.jl.git
* https://github.com/JonathanLoz/Vortice.jl.git
* https://github.com/onetonfoot/Bonsai.jl.git
* https://github.com/mcabbott/ArrayAllez.jl.git
* https://github.com/orangeSi/Jfire.jl.git
* https://github.com/chrisbrahms/Hankel.jl.git

</details>

ericphanson:
> Primarily with things being rected because there LISCENSE.md is complicated.
> Try julia's maybe?
> https://github.com/JuliaLang/julia/blob/master/LICENSE.md

```julia
julia> using LicenseCheck

julia> find_license(".")
(license_filename = "LICENSE.md", licenses_found = ["MIT"], license_file_percent_covered = 36.693548387096776)
```
would not pass, too much of the license file is unidentified.

ericphanson:
https://github.com/FluxML/Torch.jl I think is a good example of a package with a legit but complicated license which fails the simple check. It would be annoying for packages like that to require a manual merge for every version, IMO.

```julia
julia> using AnalyzeRegistry

julia> pkg = analyze_from_registry(find_packages("Torch")[1])
Package Torch:
  * repo: https://github.com/FluxML/Torch.jl.git
  * uuid: 6a2ea274-3061-11ea-0d63-ff850051a295
  * is reachable: true
  * lines of Julia code in `src`: 3809
  * lines of Julia code in `test`: 181
  * has license(s) in file: MIT, BSD-3-Clause
    * filename: LICENSE
    * OSI approved: true
  * has documentation: false
  * has tests: true
  * has continuous integration: true
    * GitHub Actions
    * GitLab Pipeline

julia> pkg.license_files
1-element Vector{NamedTuple{(:license_filename, :licenses_found, :license_file_percent_covered), Tuple{String, Vector{String}, Float64}}}:
 (license_filename = "LICENSE", licenses_found = ["MIT", "BSD-3-Clause"], license_file_percent_covered = 66.46706586826348)
```
(the problem being only 66% of the file is covered, under the 70% threshold).

giordano:
Would it be possible to have this as a warning rather than a fatal error for automerge for some time (some months)?  In this way we can encourage existing packages to have a license/`license` field in the project without stopping AutoMerge too often

ericphanson:
> Interestingly NPM advises closed source stuff to set:
> 
> ```
> "license": "UNLICENSED"
> ```
> 
> which i guess could be a useful guard if you are in the process of open sourcing something, but don't want this commit to be open sourced.
> Or if you are doing some non-open source, but source public thing like MongoDB does.

That's not allowed in General btw (policy is to require an OSI-approved license)

DilumAluthge:
Instead of 
```toml
license = "MIT"
```

What if we instead required:
```toml
license = "https://opensource.org/licenses/MIT"
```

Explicitly linking to the full text of the license on a web page hosted by the OSI makes it more explicit that you are licensing your package under the MIT license. And since you are linking to the license text, there is no ambiguity what version of the text you intended to use.

My concern with `license = "MIT"` is that this text does not in any way say that "MIT" should be interpreted according to SPDX.

Whereas if we do `license = "https://opensource.org/licenses/MIT"`, it's more clear what they meant.

And presumably doing `license = "https://opensource.org/licenses/MIT"` is no more difficult than doing `license = "MIT"`.

Parsing this `license` field on our end is really easy. There is a finite set of OSI-approved licenses, available at this link: [https://opensource.org/licenses/alphabetical](https://opensource.org/licenses/alphabetical). So we just assemble the set of all acceptable license URLs, and check that `license` is an element of that set.

oxinabox:
I would be much more down with a URL like that than a path to a file in the repo.

If we didn't restrict it to a fixed list and just retrieved it and checked it 
then that URL could even point to your extra file.
Downside there is that URLs are not persistent.

Downside of a fixed list (that would point to somewhere stable) is it feels arbitrary and capricious.
And will be a minor frustration when you think you have a good link but then get rejected.

---

I will say again though.
SPDX was good enough for NPM.
It's surely good enough for us.
They aldo don't mention in the file that the field is SPDX. Just in the docs

oxinabox:
> That's not allowed in General btw (policy is to require an OSI-approved license)

Exactly.
Which is why it would be useful to put there.
So no one accidentally registers it by mistake.
As a guard against a false positive on the LiSCENSE.md

DilumAluthge:
I would rather do a fixed list than have to fetch the contents of the URL.

There is only a finite list of OSI approved licenses.

If the OSI approves a new license and someone wants to use that license, we'll just add the new URL to our list.

Nosferican:
Just a couple comments. Having the license as part of the metadata (e.g., `Project.toml`) is preferable as part of making it machine readable. However, that does not preclude people from having the `LICENSE.md` file in the top level which is a good practice and also allows for GitHub to use Licensee to record it as part of the repository metadata in the forge. Using the SPDX code for OSI-approved licenses is the best approach for that. Might be worth considering adding a `copyright_holder` since that is also part of the license information beyond the terms characterized by the `SPDX`.

DilumAluthge:
I just don't think that this is sufficient:
```toml
license = "MIT"
```

In contrast, this is no more work on the part of the package author, and is much less ambiguous:
```toml
osi_license_url = "https://opensource.org/licenses/MIT"
```

The latter approach makes it much more clear what the intent of the author is.

Just because NPM uses the SPDX identifiers doesn't mean that we should. NPM does not have a rule that you must have an OSI-approved open source license. We do have that rule. Because we have that rule, we know that there is a finite set of licenses that we will accept.

This PR has been open for a while. Here are the steps I would suggest @ericphanson:
1. Remove the `license = "MIT"` functionality.
2. Add the `osi_license_url = "https://opensource.org/licenses/MIT"` functionality. You can just hardcode the list of OSI approved licenses from the horse's mouth: https://opensource.org/licenses/alphabetical

I am comfortable merging this PR (and deploying it onto General) as soon as those two steps are done.

DilumAluthge:
Also, most people won't need to use the `osi_license_url` field at all. As Eric posted above, there are 351 packages for which the license autodetection fails.

The General registry has 5162 packages.

So only approximately 7% of all packages will need to use the `osi_license_url` field. Everyone else can use license autodetection.

I think it's fine to have 7% of people put a URL in the `osi_license_url` field.

DilumAluthge:
>  And will be a minor frustration when you think you have a good link but then get rejected.

If this happens, then we will just manually merge the PR, and then we'll update the list of acceptable URLs in the RegistryCI.jl source code.

AutoMerge does not need to be perfect or cover every possible edge case. It just needs to cover most cases. And then the remaining cases can be manually handled.

giordano:
> This PR has been open for a while. Here are the steps I would suggest @ericphanson:
> 
> 1. Remove the license = "MIT" functionality.
> 2. Add the osi_license_url = "https://opensource.org/licenses/MIT" functionality. You can just hardcode the list of OSI approved licenses from the horse's mouth: opensource.org/licenses/alphabetical

Alternatively doing only 1 and trying to do 2 in a later pull request, since the check in `Project.toml` seems to be the most controversial part?

DilumAluthge:
> > This PR has been open for a while. Here are the steps I would suggest @ericphanson:
> > 
> > 1. Remove the license = "MIT" functionality.
> > 2. Add the osi_license_url = "https://opensource.org/licenses/MIT" functionality. You can just hardcode the list of OSI approved licenses from the horse's mouth: opensource.org/licenses/alphabetical
> 
> Alternatively doing only 1 and trying to do 2 in a later pull request, since the check in `Project.toml` seems to be the most controversial part?

Yes, I would be comfortable with this as well.

ericphanson:
I am concerned that that might make for a sudden surge in manual work without a simple solution: as soon as we turn this on, new versions will need fixing or manual merging for every package that does not pass the guidelines, which includes ones with ok licenses like Torch.jl. I was hoping we could tell those packages with licenses but who don’t pass the check “just add this field and you won’t need manual merging for every version”, or some other fix that lets them go back to automatic mode.

So I’d rather not start with the most strict option and add more options later; I’d rather start with a looser option and tighten it later. To that end, what about removing the project stuff (since that’s controversial), and to compensate, also relaxing the check in two ways?

1. Lowering the “threshold of license file recognized” requirement to zero (from an arbitrary 70%). That is, we still require that licensecheck recognizes a license in a top level file in the package, but we don’t require that licensecheck understands anything more in the file that that it has a license. (So it could be at the bottom of a readme, or have a bunch of unrecognized licenses too). This will let eg Torch.jl pass.
2. Requiring only that one of the licenses found is OSI-approved, not all of them. That will rule out some false positives like CommonMark.jl which put a CC0 license too.

I think these two requirements will cause most of the false positives, and when we have a better way to deal with false positives, we could add them back in (since I think they do have a legitimate function).

DilumAluthge:
Sure that sounds fine.

I agree that long-term, we want to add both of those requirements back in. But we can do that after we have implemented the project override feature.

DilumAluthge:
> > And will be a minor frustration when you think you have a good link but then get rejected.
> 
> If this happens, then we will just manually merge the PR, and then we'll update the list of acceptable URLs in the RegistryCI.jl source code.
> 
> AutoMerge does not need to be perfect or cover every possible edge case. It just needs to cover most cases. And then the remaining cases can be manually handled.
> 
> EDIT: Also, this will only occur when both of the following occur:
> 
>     1. The OSI approves a new license.
> 
>     2. Someone wants to use this new license for a Julia package.
> 
> 
> So it will be a very infrequent event.

Actually, I just realized that LicenseCheck.jl uses a hardcoded list of which SPDX identifiers are OSI-approved: https://github.com/ericphanson/LicenseCheck.jl/blob/main/src/OSI_LICENSES.jl

So, even with the `license = "MIT"` approach in this PR, if OSI approves a new license, we need to regenerate the hardcoded list in LicenseCheck.jl.

So that disadvantage is shared by both the `license = "MIT"` and the `osi_license_url = "https://opensource.org/licenses/MIT"` approach, right?

@ericphanson Is it correct that the list of OSI-approved licenses is hard-coded in LicenseCheck.jl?

DilumAluthge:
Why don't we have a Pkg call tomorrow to discuss the `license` field in the project file?

DilumAluthge:
@ericphanson Maybe hold off on making any changes to the PR for now? If we have a Pkg call tomorrow, and in the call it is decided that we go the `license = "MIT"` approach, then your PR will be good to go.

ericphanson:
> @ericphanson Is it correct that the list of OSI-approved licenses is hard-coded in LicenseCheck.jl?

Yep! The script to do so is in the repo as well so we can update it, but it definitely does need manual updating. Additionally, the list of licenses `licensecheck` (the underlying Go library) recognizes is shipped with the binary, i.e. we need to manually update *that* as well for new licenses (hopefully upstream updates and we just need to release a new version of our Yggdrasil binary though).

> Why don't we have a Pkg call tomorrow to discuss the license field in the project file?

Sure, I can probably attend that.

> @ericphanson Maybe hold off on making any changes to the PR for now? If we have a Pkg call tomorrow, and in the call it is decided that we go the license = "MIT" approach, then your PR will be good to go.

Sounds good, although it probably needs integration tests and some rebasing, and I might not have time to do that until this weekend anyway.

ericphanson:
Since we didn't end up having a package call, I'm thinking the next steps are to

- [ ] make the two relaxations I mention in https://github.com/JuliaRegistries/RegistryCI.jl/pull/344#issuecomment-788351425
- [ ] remove the Project.toml override
- [ ] add integration tests

Does that sound good?

DilumAluthge:
Yeah.

ericphanson:
I add some unit tests that try out the various code paths in `meets_version_has_osi_license`.

What needs to be tested by integration tests then is that the `depot_path` field gets correctly populated with a depot that we can find the package in during the guideline. So I just turned on `check_license` for Requires.jl since that passes currently and so should be getting `Pkg.add`'d successfully, so we should be able to test the codepath about populating the depot and getting the package out of the depot. Let's see if it works.

bors try

ericphanson:
bors try

ericphanson:
Ok, I think this is good to go from my end! Bors can't squash, right? So I should do some git cleaning-up.

Do the messages look OK? Too verbose?

DilumAluthge (approved):
Good to go. Yeah I would squash before you `bors merge`.

ericphanson:
bors merge

PR #348, opened 2021-02-13
Mitigate dependency confusion

GunnarFarneback:
This PR adds a dependency confusion mitigation mechanism for public registries beside General, which are particularly exposed to the vulnerability since their UUIDs are, well, public.

The idea, once this PR goes live, is that public registries who want to take part of this mitigation would submit a PR against https://github.com/JuliaRegistries/General/blob/master/.github/workflows/automerge.yml to add their repo URL to the `RegistryCI.AutoMerge.run` argument `public_registries`. When automerge runs, it will clone those repositories and check for a UUID conflict when examining a new-package request. If a conflict is found automerge is blocked, usually.

There is one exception to this rule. Assume you want to add a package to General that has previously lived in one of the checked public repositories. Obviously this will trigger a UUID conflict. However, in this case the package repository will match, which means that
1. An attacker has no way to add rouge versions.
2. Registrator wouldn't have accepted the registration request from someone unauthorized.

So the rule is that automerge is blocked for conflicting UUIDs, unless repository URL and package name match. (Same UUID with different names is no good, regardless of dependency confusion.)

There are some outstanding questions of how much time this takes and whether it gracefully handles connection problems for the public registries.

Cc: @timholy


oxinabox:
A slight generalization would be if we could hook this up either to a registry repo URL,
or just a URL that contains a plain text file with one UUID per line.
That way private registries would have the option of publishing their list of UUIDs, and we could protect them.

GunnarFarneback:
I agree that some blacklisting solution also for private registries would be desirable but I'm not convinced that it should be plain text and for this PR the focus is to give the more exposed public registries a tool to work with. However, a private registry which really wants to take part in this scheme could make a public dummy registry, which wouldn't have to be much more than a list of UUIDs.

GunnarFarneback:
For integration testing it should work to set up a "public registry" in a temporary directory and point to it with a `file://` URL.

GunnarFarneback:
bors try

GunnarFarneback:
bors try

GunnarFarneback:
bors try

GunnarFarneback:
bors try

GunnarFarneback:
bors try

GunnarFarneback:
bors try

GunnarFarneback:
bors try

src/AutoMerge/dependency_confusion.jl
```diff
@@ -0,0 +1,50 @@
...
+    Guideline("No UUID conflict with other registries.",
+              data -> has_no_dependency_confusion(data.pkg,
+                                                  data.registry_head,
+                                                  data.public_registries))
+
+# TODO: Needs a strategy to handle connection failures for the public
+# registries. Preferably they should also be cloned only once and then
+# just updated to mitigate the effect of them being temporarily
+# offline. This could be implemented with the help of the Scratch
+# package, but requires Julia >= 1.5.
```
ericphanson, line 14: I'd be in favor of bumping the minimum version to Julia 1.5 if it helps here; the syntax benefits would be nice too (no more `x=x` in keyword arguments). And since RegistryCI is usually run by itself in CI (and is a leaf of the dependency tree), one doesn't need to upgrade their whole stack to use the new version.

src/AutoMerge/dependency_confusion.jl
```diff
@@ -0,0 +1,50 @@
...
+function has_no_dependency_confusion(pkg, registry_head, public_registries)
+    # We know the name of this package but not its uuid. Look it up in
+    # the registry that includes the current PR.
+    packages = Pkg.TOML.parsefile(joinpath(registry_head, "Registry.toml"))["packages"]
```
ericphanson, line 18: Not sure it really matters, but we already depend on TOML.jl anyway:
```suggestion
    packages = TOML.parsefile(joinpath(registry_head, "Registry.toml"))["packages"]
```

src/AutoMerge/dependency_confusion.jl
```diff
@@ -0,0 +1,50 @@
...
+    filter!(packages) do (key, value)
+        value["name"] == pkg
+    end
+    # For Julia >= 1.4 this can be simplified with the `only` function.
+    always_assert(length(packages) == 1)
+    uuid = first(keys(packages))
+    # Also need to find out the package repository.
+    package_repo = Pkg.TOML.parsefile(joinpath(registry_head, packages[uuid]["path"], "Package.toml"))["repo"]
```
ericphanson, line 26:
```suggestion
    package_repo = TOML.parsefile(joinpath(registry_head, packages[uuid]["path"], "Package.toml"))["repo"]
```

src/AutoMerge/dependency_confusion.jl
```diff
@@ -0,0 +1,50 @@
...
+    for repo in public_registries
+        registry = clone_repo(repo)
+        registry_toml = Pkg.TOML.parsefile(joinpath(registry, "Registry.toml"))
```
ericphanson, line 29:
```suggestion
        registry_toml = TOML.parsefile(joinpath(registry, "Registry.toml"))
```

src/AutoMerge/dependency_confusion.jl
```diff
@@ -0,0 +1,50 @@
...
+                             packages[uuid]["name"], " in registry ",
+                             registry_toml["name"], " at $repo.\n",
+                             "This could be a dependency confusion attack.")
+            # Conflict detected. This is benign if the package name
+            # *and* the package URL matches.
+            if packages[uuid]["name"] != pkg
+                return false, message
+            end
+            package_path = packages[uuid]["path"]
+            other_package_repo = Pkg.TOML.parsefile(joinpath(registry, package_path, "Package.toml"))["repo"]
```
ericphanson, line 42:
```suggestion
            other_package_repo = TOML.parsefile(joinpath(registry, package_path, "Package.toml"))["repo"]
```

src/AutoMerge/dependency_confusion.jl
ericphanson: a related concern is if an external registry gets corrupted and has an invalid TOML file, say. Then RegistryCI would throw everytime it gets to this check. I wonder if we should use `TOML.tryparsefile` and fail the check if we get an error, so at least RegistryCI doesn't crash in that event, but we just fail this check each time. That would still be pretty disruptive though.

StefanKarpinski:
It seems like we should do this as well as some other mitigations — defense in depth and all that. I agree that if we have a mechanism for private repos to publish this that it should be salted and hashed. That can be done separately. We can develop that in the future. Good to let public non-General registries have some protection ASAP and being able to register themselves here is simple and effective.

src/AutoMerge/dependency_confusion.jl
GunnarFarneback: It can definitely be done with Scratch, but maybe that's a secondary concern.

Invalid TOML files or other registry irregularities is a real risk. The easiest protection would be a `try/catch`, but that's rather heavyhanded.

src/AutoMerge/dependency_confusion.jl
GunnarFarneback: I just followed some existing code but would never mind shortening the code. :-)

src/AutoMerge/dependency_confusion.jl
ericphanson: I think a try/catch around that would be OK; it seems like that code should only fail if there are issues cloning and parsing the registry. And if we fail the check in the event of an error, we won't be sweeping any issues under the rug, since automerge will fail.

DilumAluthge:
Eric and Gunnar, once you're satisfied with this, go ahead and merge it.

src/AutoMerge/dependency_confusion.jl
GunnarFarneback: It's slightly unfair to the affected PRs since it's in no way their fault (well, unless they have made a DOS attack against the registry to try to disable the check...), but I agree that getting attention to the issue is of higher value.

Does someone know how  `clone_repo` behaves when there are network problems?

DilumAluthge:
And then register a new release of RegistryCI so that we can deploy this.

src/AutoMerge/dependency_confusion.jl
ericphanson: yeah, agreed it's not great for the affected PRs... it seems too risky to pass the check if there's an error though, since then an attacker could even do something like add a new public registry they control, then invalidate the TOML or cause network issues in that registry, and then get their package in during the auto-pass window before the problem is fixed.

GunnarFarneback: I've added a `try/catch` with an explanatory message.

GunnarFarneback:
Before merging it would be good to squash it, there's lots of nonsensical history from the bors iterations.

GunnarFarneback:
bors try

ericphanson (approved):
LGTM, let’s merge tomorrow so we’re awake to handle any issues that arise

GunnarFarneback:
bors merge

PR #359, opened 2021-03-12
Codecov: require that the project coverage always stays above 70%

DilumAluthge:


DilumAluthge:
bors try

DilumAluthge:
bors try

GunnarFarneback:
I don't know what levels we are at and although it's good to follow up on coverage I'm suspicious about coverage requirements. First because I've found the code coverage to be rather unreliable (I'd be happy if that is better on 1.6) and second because it can go down in ways you don't really control. For example, assume you find a nice code simplification that allows replacing 50 lines of code with just 5. If those were covered both before and after the change, you have lost 45 covered lines and your coverage ratio will drop, possibly substantially and below some critical threshold. Still it's a great change to get rid of 45 lines of code that weren't actually needed.

DilumAluthge:
> For example, assume you find a nice code simplification that allows replacing 50 lines of code with just 5. If those were covered both before and after the change, you have lost 45 covered lines and your coverage ratio will drop, possibly substantially and below some critical threshold. Still it's a great change to get rid of 45 lines of code that weren't actually needed.

Yeah, if the coverage ratio drops due to something like this, we would obviously still want to keep a change that removes a lot of unnecessary code. In this case, we can just edit the `codecov.yml` file and lower the threshold.

The purpose of this PR is really just to make sure that we are keeping an eye on coverage. Right now, we don't even notice coverage changes.

We can definitely lower the threshold as needed.

DilumAluthge:
bors try

DilumAluthge:
bors merge

ericphanson:
What about just turning on the comment, instead of a threshold? So we can see how a PR affects coverage

DilumAluthge:
> What about just turning on the comment, instead of a threshold? So we can see how a PR affects coverage

I don't know how to get that to work with Bors :(

Issue #349, opened 2021-02-22
Automerge doesn't catch incorrect `subdir` parameters


ericphanson:
hence https://github.com/JuliaRegistries/General/pull/30577. We could check this by `dev`ing the package (instead or in addition to `add`ing it):
```julia
(jl_4zbq4I) pkg> dev YisyAIFramework
     Cloning git-repo `https://github.com/SkyWorld117/YisyAIFramework.jl.git`
   Resolving package versions...
ERROR: expected package `YisyAIFramework [1833fadb]` to exist at path `/Users/eph/.julia/dev/YisyAIFramework/https:/github.com/SkyWorld117/YisyAIFramework.jl`
```
If `dev` doesn't throw, then at least it is a valid directory, though we probably should go one step further and check that the tree hash of the contents at that path matches the one in the registry. If anyone has any other ideas I'd be happy to hear it.

DilumAluthge:
Yeah we should do both `Pkg.add` and `Pkg.dev` to make sure that both work.

DilumAluthge:
> If `dev` doesn't throw, then at least it is a valid directory, though we probably should go one step further and check that the tree hash of the contents at that path matches the one in the registry.

Also agree with this.

GunnarFarneback:
> If dev doesn't throw, then at least it is a valid directory, though we probably should go one step further and check that the tree hash of the contents at that path matches the one in the registry. If anyone has any other ideas I'd be happy to hear it.

It's not quite that easy to go deeper since the registered version wasn't necessarily from the default branch and the latest commit, or there may have been more commits added between the registration and the automerge run. To more properly check that the tree hash is in the right place you need a git clone and do something equivalent to `git rev-list --objects --all | grep <TREE_HASH>` and see that you get the expected subdirectory out of that.

GunnarFarneback:
If Registrator passed on which commit hash it based the registration on it would be much easier to check: `git rev-parse <COMMIT_HASH>:<subdir>`.

DilumAluthge:
> If Registrator passed on which commit hash it based the registration on it would be much easier to check: `git rev-parse <COMMIT_HASH>:<subdir>`.

That should be available in the Registrator PR, right?

GunnarFarneback:
Right, it looks like it is. Then all that's needed is git clone. :-)

DilumAluthge:
Like you said on Slack, that Git clone can be useful for other purposes as well.

[...]

PR #360, opened 2021-03-12
Test that we can `Pkg.dev` the package (in addition to being able to `Pkg.add` the package)

DilumAluthge:
This is related to issue #349.

This does not fix issue 349. In order to fix issue 349, we also need to do this: (from [this comment](https://github.com/JuliaRegistries/RegistryCI.jl/issues/349#issue-813011226))
> If dev doesn't throw, then at least it is a valid directory, though we probably should go one step further and check that the tree hash of the contents at that path matches the one in the registry. If anyone has any other ideas I'd be happy to hear it.

But this is a good start.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>", "contributors"]
-version = "6.7.5"
+version = "6.8.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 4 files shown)


DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
So... there is a big problem with this approach.

When you do `Pkg.develop("MyPackageName")`, it grabs the latest commit from the default branch of the package repository.

But, it's not uncommon for people to treat the default branch of their repo as an unstable branch. Therefore, the code in the default branch of a package might not be installable, which is perfectly fine.

But that means that `Pkg.develop("MyPackageName")` might fail just because the default branch of the package is not installable.

So I don't think this is the correct way to deal with #349. I think to fix issue 349, we'll need to clone the package and check that the tree hash at the given subdir matches the tree hash in the Registrator PR.

cc: @ericphanson @GunnarFarneback

PR #361, opened 2021-03-14
Better error message for checking if the UUID of the dependencies is in the registry

SebastianM-C:
Credit to @ericphanson 

Rewrite dependency uuid check so that it gives a better error message when an UUID in `depuuids` is not in `alluuids`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>", "contributors"]
-version = "6.7.5"
+version = "6.7.6"
 
 [deps]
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Thanks @SebastianM-C!

I've merged #362, which is a tweaked version of this PR. I think that solves the issue - let me know if there are any remaining issues.

SebastianM-C:
Thanks for the quick reply!
One possible enhancement would be to use a more general error message such that it accommodates the case where RegistryCI.jl is used in a custom registry (for example by mentioning the names or urls of the registries involved).

DilumAluthge:
Hmm, well in that case, you know what the external registries are, right? You passed them in the `registry_deps` argument.

SebastianM-C:
Yeah, right. 🤔 
I have a question regarding that usage. In a custom registry should the General registry be included in `registry_deps` if there are packages with dependencies in there?

DilumAluthge:
Yeah that's exactly what the `registry_deps` keyword argument is supposed to do.

I think you've found a bug. Bug fix here: #367

PR #364, opened 2021-03-14
AutoMerge should not crash if the registry consistency tests fail

DilumAluthge:
Currently, if the registry tests fail, AutoMerge crashes ungracefully. This is not very helpful.

Instead, this PR makes the registry consistency tests into a regular AutoMerge check just like everything else.


DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
bors merge

PR #366, opened 2021-03-14
fix indentation in pkg name similarity list

ericphanson:
I tested this by

```julia
using RegistryCI
using RegistryCI.AutoMerge
using RegistryCI.AutoMerge: meets_distance_check, get_all_non_jll_package_names
registry_master = joinpath(homedir(), ".julia", "registries", "General")
other_packages = get_all_non_jll_package_names(registry_master)
pkg_name = "C"
meets_distance_check(pkg_name, other_packages)[2] |> println
```

then editing the comment in https://github.com/JuliaRegistries/General/pull/31746, copy pasting the result, and clicking preview to check that it works properly even with the spoiler tag:

Screenshot 2021-03-14 at 18 40 41

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>", "contributors"]
-version = "6.8.4"
+version = "6.8.5"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Yeah the only weird part is that in the spoiler tag case, all of a sudden, there is a lot of space between each bulleted item in the top-level bulleted list.

Probably not a big deal though.

DilumAluthge:
Needs a rebase on master I think.

DilumAluthge (approved):

DilumAluthge:
You may want to squash right before you `bors merge`.

ericphanson:
the combination of version vigilante + bors not being able to squash merge is pretty annoying. I think I spent as much time dealing with the version number conflicts and squashing/rebasing as actually writing/testing the changes (which wasn't that much time since it's a small change, but that's exactly the annoying thing-- small changes should be easy and quick).

> Yeah the only weird part is that in the spoiler tag case, all of a sudden, there is a lot of space between each bulleted item in the top-level bulleted list.

That is weird, but to me seems like a github markdown bug as much as anything else..

bors merge

PR #367, opened 2021-03-14
Fix the `load_registry_dep_uuids` function

DilumAluthge:


DilumAluthge:
cc: @SebastianM-C

DilumAluthge:
bors try

SebastianM-C:
The fix seems to work on my side: https://github.com/ctp-fpub/JuliaRegistry/runs/2107565108?check_suite_focus=true

The failures on 1.5.3 and 1.6 may be because I only updated the Manifest.toml and not the old one.

DilumAluthge:
Yeah, you'd also need to update the manifests that the 1.5 and 1.6 jobs use.

Anyway, I'll merge this soon, and immediately after that I'll register a new release of RegistryCI.

SebastianM-C:
Thanks!

DilumAluthge:
Thanks for finding and reporting this bug!

I've added a test as well, so we can make sure the bug doesn't pop up again in the future.

src/registry_testing.jl
```diff
@@ -15,6 +15,33 @@ else
...
+    end
+    for fieldvalue in all_fieldvalues
+        for registry_deps_name in registry_deps_names
+            if strip(fieldvalue) == strip(registry_deps_name)
+                return true
+            end
+            if strip(fieldvalue) == strip("$(registry_deps_name).git")
+                return true
+            end
+            if strip("$(fieldvalue).git") == strip(registry_deps_name)
```
SebastianM-C, line 37: So the bug was caused because I did not include the full link with `.git` at the end?

SebastianM-C (approved):

src/registry_testing.jl
DilumAluthge: Nope. The bug was due to a change in the Pkg internals. Because we didn't have any unit tests covering this code, we did not detect the change.

I took this opportunity to also add some code that takes care of links both with and without `.git` at the end. But the main bug was unrelated to that.

DilumAluthge:
bors merge

src/registry_testing.jl
SebastianM-C: Thanks

PR #372, opened 2021-03-15
Update VersionVigilante_bors.yml

ericphanson:
Add `paths-ignore` options. Rationale: version bumps are only needed for code changes, i.e. things that affect what happens when you do `using RegistryCI` and onwards.

### DIFF
--- a/.github/workflows/VersionVigilante_bors.yml
+++ b/.github/workflows/VersionVigilante_bors.yml
@@ -5,6 +5,11 @@ on:
     branches:
       - staging
       - trying
+    paths-ignore:
+      - '*.md'
+      - 'bors.toml'
+      - 'codecov.toml'
+      - '.github/*'
 
 jobs:
   VersionVigilante:
--- a/.github/workflows/VersionVigilante_bors.yml
+++ b/.github/workflows/VersionVigilante_bors.yml
@@ -9,7 +9,7 @@ on:
       - '*.md'
       - 'bors.toml'
       - 'codecov.toml'
-      - '.github/*'
+      - '.github/**'
 
 jobs:
   VersionVigilante:


.github/workflows/VersionVigilante_bors.yml
```diff
@@ -5,6 +5,11 @@ on:
     branches:
       - staging
       - trying
+    paths-ignore:
+      - '*.md'
+      - 'bors.toml'
+      - 'codecov.toml'
+      - '.github/*'
```
ericphanson, line 12:
```suggestion
      - '.github/**'
```
since I want to match files in the workflows subdir too (ref https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags).

DilumAluthge:
I don't think this will work. Bors needs a green (passing) status. IIUC, this will just cause VersionVigilante not to run.

DilumAluthge:
You'd need to be able to return a green status if the path is in the paths-ignore.

ericphanson:
Ah darn.. Maybe we can add a conditional to the `job` but not the workflow? ie via https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif. So then the workflow will run just not the particular VV job within it?

DilumAluthge:
Yeah that would probably work.

ericphanson:
I don't see an easy way to get the paths info via the Github API; we could shell out to git and figure out the paths, but that seems like a lot of work. But I see that Bors allows a syntax for matching one of multiple commit statuses: https://bors.tech/documentation/. So maybe we can have a second workflow with a `paths` condition with these same paths that just outputs a passing indicator. Then only one of the two will run, and Bors will get its pass.

ericphanson:
Let me test this... maybe it just needs to not have a failing status?

bors try

ericphanson:
bors try

DilumAluthge:
It's in the queue. It'll start once the "try" I have running on #373 finishes.

DilumAluthge:
> I don't see an easy way to get the paths info via the Github API; we could shell out to git and figure out the paths, but that seems like a lot of work. But I see that Bors allows a syntax for matching one of multiple commit statuses: https://bors.tech/documentation/. So maybe we can have a second workflow with a `paths` condition with these same paths that just outputs a passing indicator. Then only one of the two will run, and Bors will get its pass.

This sounds promising.

Let me land #373 before you work on this approach, since otherwise you might have merge conflicts.

ericphanson:
> maybe it just needs to not have a failing status?

Nope, looks like it needs the passing status.

It would be nice to be able to say something like “bors merge patch” or “bors merge minor” and it could do the bump if needed. Anyway that’s probably well beyond what bors can do.

DilumAluthge:
> It would be nice to be able to say something like “bors merge patch” or “bors merge minor” and it could do the bump if needed. Anyway that’s probably well beyond what bors can do.

Yeah I've wanted this kind of feature for a while now.

ericphanson:
closing this since it needs a new approach (like https://github.com/JuliaRegistries/RegistryCI.jl/pull/372#issuecomment-799623268) and not sure when I'll have time/will to work on it

PR #379, opened 2021-04-18
clone and use tree hash to get code instead of `Pkg.add`

### LINKED ISSUE #349: Automerge doesn't catch incorrect `subdir` parameters

ericphanson:
* adds a non-optional `guideline_code_can_be_downloaded` check that cross-checks the tree hash from the PR's Versions.toml against the commit hash from the PR message body, and downloads the code to a `pkg_code_path` in the `GitHubAutoMergeData` object (replacing the `depot_path` that used to be there)
* The package can be `Pkg.add`'d and `import`ed steps now use a tempdir instead of the `depot_path` since we don't need to reuse the code anymore
* The license check now looks in the `pkg_code_path` to get the code
* breaking version number bump due to the requirement that PR bodies include the commit hash

closes #349


src/AutoMerge/util.jl
```diff
@@ -26,6 +22,48 @@ function _clone_repo_into_dir(url::AbstractString, repo_dir)
...
+function load_files_from_url_and_tree_hash(destination::String, url::String, tree_hash::String)
+    pkg_clone_dir = mktempdir()
+    clone_success = try
+        _clone_repo_into_dir(url, pkg_clone_dir)
+        true
+    catch e
+        @error "Cloning $url failed"
+        false
+    end
+    Tar.extract(Cmd(`git archive $tree_hash`; dir=pkg_clone_dir), destination)
```
ericphanson, line 42: The fact that `Tar.extract` takes a `Cmd` as input is really nice!

DilumAluthge: @StefanKarpinski (for both Tar.jl and `Cmd` 😂)

GunnarFarneback:
Looks good so far.

ericphanson:
Added the hash code! Let's see what bors thinks

@bors try

DilumAluthge:
You don't need the `@` before `bors`.

ericphanson:
bors try

ericphanson:
Oops, guess `Base.run` instead of `run` is needed afterall 😅. That was a confusing error message!

bors try

ericphanson:
bors try

ericphanson:
It looks like the tests might be failing here because we don't populate the PR bodies for the integration tests, e.g. https://github.com/bcbi-test/automerge-integration-test-repo/pull/7550

ericphanson:
I think we could do something like make the download code check enable-able by an argument, then disable it for most of the integration tests (since many correspond to made-up versions anyway), but enable it for one that corresponds to actual commit, and pass in the correct commit in the PR body for that one. I'll try that...

DilumAluthge:
> It looks like the tests might be failing here because we don't populate the PR bodies for the integration tests, e.g. [bcbi-test/automerge-integration-test-repo#7550](https://github.com/bcbi-test/automerge-integration-test-repo/pull/7550)

Let's add PR bodies for all the integration tests. The point of the integration tests is to replicate the real-world conditions as closely as possible.

DilumAluthge:
> I think we could do something like make the download code check enable-able by an argument, then disable it for most of the integration tests (since many correspond to made-up versions anyway), but enable it for one that corresponds to actual commit, and pass in the correct commit in the PR body for that one. I'll try that...

Personally I'd rather not special case anything. Can we just do the "download code check" for all of the integration tests? IIRC, the Git tree SHAs in the integration tests all correspond to real commit in those repos.

ericphanson:
> Let's add PR bodies for all the integration tests. The point of the integration tests is to replicate the real-world conditions as closely as possible.

I'm not sure it would add that much beyond what the tests I just added do, but it's a lot of configuration to add. With those we should already cover both cases (right commit and wrong commit).

bors try

ericphanson:
> Personally I'd rather not special case anything. Can we just do the "download code check" for all of the integration tests? IIRC, the Git tree SHAs in the integration tests all correspond to real commit in those repos.

Sorry, I didn't see this before my last reply. I see, just the names and versions are made up then. That should be possible then, but I'm not sure it tests that much more than doing it once? It seems like we'd just be re-checking that the same commit + subdir pair corresponds to the same git-tree-sha many times, since the git-tree-sha's are reused throughout the tests.

bors try

DilumAluthge:
My main concern was that it sounds like the integration tests don't reflect the way things are in production. So I figure if we just add PR bodies, then the integration tests run exactly the same way things do in production.

I don't mind running the same check over and over. What worries me is if the integration tests don't line up with production.

ericphanson:
Ok, so I added the hashes so we test unconditionally. I haven't removed the `check_code_downloadable` option (though it defaults to true), though I think maybe I should? Wasn't sure if there could be some weird configuration problems on private registries and it would be nice to be able to disable it in that case, though in that case we should probably want to fix whatever the issue is anyway.

bors try

DilumAluthge:
> Ok, so I added the hashes so we test unconditionally. I haven't removed the `check_code_downloadable` option (though it defaults to true), though I think maybe I should? Wasn't sure if there could be some weird configuration problems on private registries and it would be nice to be able to disable it in that case, though in that case we should probably want to fix whatever the issue is anyway.
> 
> bors try

Good question. It occurs to me that maybe some private registry setups might have firewalls that block Git or something?

So, how does this sound: leave the kwarg, but don't document, so it's private. Then we can remove it later if we want, without needing a breaking release.

ericphanson:
Sounds good! This will be my last attempt tonight but I'll try to get back at it tomorrow.

bors try

DilumAluthge:
Actually, I think we should do the following:
1. Remove the `check_code_downloadable` keyword argument.
2. Bump the major version number

My justification for the latter: right now, we only require that PRs have the package name and the version number. So people in private registries may be using AutoMerge with pull requests that are not opened by Registrator and thus do not have the commit in the PR body. This might make sense in a private registry setup, since you can give users direct merge permissions on the registry.

So then this is technically a breaking change since we now impose an additional constraint on the formatting of PRs.

DilumAluthge:
You can bump the major version number and put `-DEV` at the end. That way, if we have more breaking changes to make in the short term, we can make them all in a single breaking release.

GunnarFarneback:
> Good question. It occurs to me that maybe some private registry setups might have firewalls that block Git or something?

Trying to run automerge with a setup that blocks `git` would be a bit weird; the registry is typically stored in a `git` repository after all. A more likely complication is that a CI job running automerge might not be configured with sufficient keys or other credentials to access all package repos.

ericphanson:
bors try

ericphanson:
I don't know if it would be better or worse or no different for auth, but one option is we could switch the commandline invocations of `git` to https://github.com/JuliaVersionControl/Git.jl if we were willing to require Julia 1.6+.

ericphanson:
bors try

ericphanson:
Ok, should be good to do! I squashed the commits here.

bors try

DilumAluthge (approved):
Good to merge, I think.

GunnarFarneback (approved):

ericphanson:
bors merge

Issue #382, opened 2021-04-27
dry-run/staging functionality


ericphanson:
It would be helpful to have a dry-run/staging mode where RegistryCI does not:

1. Create any commit statuses
2. Create or update any PR comments
3. Merge any pull requests

but otherwise functions by reading PRs and emitting logs and throwing an error in the end if the automerge guidelines are not met.

Required for https://github.com/JuliaRegistries/General/issues/35514

PR #383, opened 2021-04-27
add `read_only` mode

ericphanson:
closes https://github.com/JuliaRegistries/RegistryCI.jl/issues/382

I thought this would be way harder! But also I could've missed something.

I wonder if we can test this by adding a github workflow with read-only permissions like we talked about in https://github.com/JuliaRegistries/General/issues/35514, and then run the regular integration tests on that except with `read_only` set to true. What do you think?

### DIFF
--- a/src/AutoMerge/automerge_comment.jl
+++ b/src/AutoMerge/automerge_comment.jl
@@ -1,5 +1,9 @@
 function update_automerge_comment!(data::GitHubAutoMergeData,
                                    body::AbstractString)::Nothing
+    if data.read_only
+        @info "`read_only` mode; skipping updating automerge comment."
+        return nothing
+    end
     api = data.api
     repo = data.registry
     pr = data.pr
--- a/src/AutoMerge/cron.jl
+++ b/src/AutoMerge/cron.jl
@@ -175,7 +175,8 @@ function cron_or_api_build(api::GitHub.GitHubAPI,
                            new_jll_version_waiting_period,
                            whoami::String,
                            all_statuses::AbstractVector{<:AbstractString},
-                           all_check_runs::AbstractVector{<:AbstractString})
+                           all_check_runs::AbstractVector{<:AbstractString},
+                           read_only::Bool)
     # first, get a list of ALL open pull requests on this repository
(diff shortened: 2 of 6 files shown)


DilumAluthge:
> I wonder if we can test this by adding a github workflow with read-only permissions like we talked about in [JuliaRegistries/General#35514](https://github.com/JuliaRegistries/General/issues/35514), and then run the regular integration tests on that except with `read_only` set to true. What do you think?

Yeah. I'm about to open a PR demo-ing the read-only token. Once I open that, want to take a look and see if you can figure out how to repurpose it?

DilumAluthge:
It's going to be tricky for you to figure out how to do integration tests with a read-only token.

Because the mechanism I use in https://github.com/JuliaRegistries/General/pull/35519 to generate a read-only token is a specific mechanism (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions) that only works for the `GITHUB_TOKEN` supplied automatically by GitHub.

But in the integration tests, I don't actually use the `GITHUB_TOKEN`. I use a PAT that I generated for my bot user.

ericphanson:
Ah ok. Maybe we can make a new PAT with read-only perms then?

DilumAluthge:
I think as a first pass, let's just leave this untested? Since we will be testing it on General. And then we can add tests to this later. (Can you open an issue so we don't forget?)

ericphanson:
Ah ok, that's fine. Let's wait until tomorrow to merge this though; it's late here and it would be good to get @GunnarFarneback's review if possible too.

DilumAluthge:
While waiting for Gunnar's review, let's make sure CI passes.

bors try

ericphanson:
bors try

DilumAluthge (approved):

GunnarFarneback:
I haven't checked if it covers all possible outputs but it certainly makes the right kind of changes and what's there looks right.

GunnarFarneback (approved):

ericphanson:
Alright, let's give it a go! If we run it with a read-only token we should get errors if I missed something here.

bors merge

PR #386, opened 2021-04-28
Fix commit regex and test it more robustly

DilumAluthge:


DilumAluthge:
I think I've figured out the issue. The integration test PR bodies are not quite the same as the real life PR bodies. So I think the regex passes on integration test PR bodies, but fails on real life PR bodies. 

This test should help catch that.

bors try

ericphanson:
Yeah, it looks like if I replace the `^` and `$` with `\n` then it works:
```julia
julia> pr_body = """
              - Registering package: TiffImages
              - Repository: https://github.com/tlnagy/TiffImages.jl
              - Created by: @tlnagy
              - Version: v0.3.2
              - Commit: 6dec1188f41ae00243e8273bd1779e661ffdba48
              - Reviewed by: @tlnagy
              - Reference: https://github.com/tlnagy/TiffImages.jl/commit/6dec1188f41ae00243e8273bd1779e661ffdba48#commitcomment-50137614
              <!-- bf0c69308befbd3ccf2cc956ac8a46712550b79fc9bfb5e4edf8f833f05f4c18b06eddad8845b45beb9f45c2b8020dd61b2f40df05dcabcc5ab0dc8f0e472a2aca7229636a59749a3b16f9aecd340f462a63ac2000ca9f6c836c001dd613d35be874e702157f4d5e9851574b75787e4780854087852c83a2995c7d8d6f4bebfad42a122576b934846df858f01e3ce71dbda9bf0f1e8c02e20869e34fb0a7fb0c3dc441dbbd66d6eb9fa0654d28e68010470cf571243f3df4d665a95139e85461cd153bed19f404d8a9f48a39c1e062412fded3ed4f569f060f790ed8205cfcac -->
              """;

julia> match(r"\n- Commit: (\w*)\n", pr_body)
RegexMatch("\n- Commit: 6dec1188f41ae00243e8273bd1779e661ffdba48\n", 1="6dec1188f41ae00243e8273bd1779e661ffdba48")
```

ericphanson:
Yeah, I think that's definitely the issue. I thought `^` and `$` meant start/end of the line, not the whole string, and my test wasn't correct enough to catch the difference. Do you mind fixing the regex in this PR? Then with the better test we should be good to go.

DilumAluthge:
But then it won't match if the PR body only has one line 😭 

What if you do this? Then it should definitely always match, whether the PR body is one line or multiple lines.
```julia 
match(r"\n- Commit: (\w*)\n", string("\n", pr_body, "\n"))
```

ericphanson:
Hm, sure that would be fine, or we could just do `r"- Commit: (\w*)"` and be a bit looser. It seems weird that that would exist and not be the thing we are trying to find.

DilumAluthge:
> I thought `^` and `$` meant start/end of the line, not the whole string, and my test wasn't correct enough to catch the difference.

I've been bitten by this before. I use https://regex101.com/, and by default, when you use https://regex101.com/, `^` and `$` do correspond to the start/end of the line, because they interpret each line as a different test string.

ericphanson:
> I've been bitten by this before. I use https://regex101.com/, and by default, when you use https://regex101.com/, ^ and $ do correspond to the start/end of the line, because they interpret each line as a different test string.

Ahh... Yeah I think I was using regex101 too. Good to know!

DilumAluthge:
Regex101 is great, you just have to keep in mind that by default they treat each line as a separate test string.

I think you can toggle that behavior though.

DilumAluthge:
bors try

ericphanson (approved):

DilumAluthge:
bors merge

DilumAluthge:
@ericphanson Can you turn staging back on?

PR #388, opened 2021-04-28
Regex improvements and regex tests

DilumAluthge:


DilumAluthge:
bors try

ericphanson:
it's starting to get more complicated 😅 . Maybe we can add unit tests for this regex? Could just check that `RegistryCI.AutoMerge.commit_regex` matches the right things and misses other things

DilumAluthge:
Good call, I'll add some tests.

DilumAluthge:
bors merge

PR #393, opened 2021-05-08
Add some checks to make sure that the regex docs never get out of sync

DilumAluthge:
With this PR, we:
1. Implement the DRY (don't repeat yourself) principle. For example, the regexes are only defined in one place.
2. Make sure that the regex documentation never gets out of sync with the source code.


DilumAluthge:
bors merge

PR #395, opened 2021-05-10
Run the integration tests on both Julia v1 and Julia nightly

DilumAluthge:


DilumAluthge:
Hmmm. Currently we run the integration tests on one version of Julia, and it takes approx. 14-16 minutes. If we run the integration tests on two versions of Julia, it will take approx 28-32 minutes. (We can't run integration tests in parallel because we'll hit the GitHub API rate limits.)

I'm not sure that it's worth increasing the CI time to be 30 minutes per PR.

PR #401, opened 2021-05-11
update docs

Roger-luo:
I updated & added some of the docs, but there are certain options I'm not sure what is that, so probably someone could help me to finish this PR.

PR #402, opened 2021-05-11
Delete remember_to_update_registryci.yml

ericphanson:
Not used anymore, and the replacement is added in #401


ericphanson:
I bumped the version so we can tag after this and get the stable docs updated.

bors merge

PR #403, opened 2021-05-12
Put all badges in a table

ericphanson:
I thought it looked a little nicer to put them all in a table together instead of some in a table and some above. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,26 @@
 # RegistryCI.jl
 
-|                         |                                                               |                   |                                                                                         |
-|------------------------|---------------------------------------------------------------|-------------------|-----------------------------------------------------------------------------------------|
-| Documentation (stable) | [![Documentation (stable)][docs-stable-img]][docs-stable-url] | Unit Tests        | [![Continuous Integration (Unit Tests)][ci-unit-img]][ci-unit-url]                      |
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Looks kind of strange with an empty header row. But that's only a GitHub-Flavored Markdown limitation, right?

Maybe use HTML for the table instead? I think if you do the table in HTML, GitHub will let you have a table without a header row.

E.g. this:
```html
<table align="center">
    <tr>
        <td align="center">some text</td>
        <td align="center">Some other text</td>
        <td align="center">More text</td>
    </tr>
    <tr>
        <td align="center">some text</td>
        <td align="center">Some other text 2</td>
        <td align="center">More text 2</td>
    </tr>
</table>
```

Produces this:

<table align="center">
    <tr>
        <td align="center">some text</td>
        <td align="center">Some other text</td>
        <td align="center">More text</td>
    </tr>
    <tr>
        <td align="center">some text</td>
        <td align="center">Some other text 2</td>
        <td align="center">More text 2</td>
    </tr>
</table>

ericphanson:
Ah, I think the markdown links don't work inside the HTML; need explicit <a>s and all

DilumAluthge:
What about something like this? https://github.com/JuliaLang/SuiteSparse.jl/blob/master/README.md

PR #404, opened 2021-05-14
README: remove the Bors badge, and put all other badges into the table

DilumAluthge:
Closes #403 

The Bors badge is static (i.e. the content never changes), so I went ahead and removed it.

I put everything else into the table. The docs badges are not very big, so I put both of them into the same cell.


ericphanson (approved):

DilumAluthge:
bors merge

PR #407, opened 2021-05-17
Merge new-package and new-version

GunnarFarneback:
There was basically no difference between `new-package.jl` and `new-version.jl` anymore, except for package/version in some output strings. This PR merges them into `pull_requests.jl`. It also removes an intermediate method of `pull_request_build`, which didn't provide any visible value.


docs/src/guidelines.md
```diff
@@ -24,7 +25,8 @@ function guidelines_to_markdown_output(guidelines_function::Function)
     return output_markdown
 end
 
-const guidelines_function = RegistryCI.AutoMerge.get_automerge_guidelines_new_packages
+const guidelines_function = RegistryCI.AutoMerge.get_automerge_guidelines
+const registration_type = RegistryCI.AutoMerge.NewPackage()
```
ericphanson, line 29: I didn't know you could define constants after they are used (I knew you could with functions) but seems like it works!

src/AutoMerge/util.jl
```diff
@@ -111,7 +111,8 @@ end
 
 function comment_text_pass(::NewVersion,
                            suggest_onepointzero::Bool,
-                           version::VersionNumber)
+                           version::VersionNumber,
+                           is_jll::Bool)
```
ericphanson, line 115: this new argument doesn't seem to be used

GunnarFarneback: It's not but it's needed to allow a uniform call to both methods.

ericphanson: https://github.com/JuliaRegistries/RegistryCI.jl/pull/407/files#diff-642068d5927ce10f0fca27aa4cc347618d5e8904fff6f21fcb2a542faa34cda3R210-R213 vs https://github.com/JuliaRegistries/RegistryCI.jl/pull/407/files#diff-642068d5927ce10f0fca27aa4cc347618d5e8904fff6f21fcb2a542faa34cda3R199-R202 look different still, where are they uniform?

GunnarFarneback: Those are different functions. It's the two methods of `comment_text_pass` that have been harmonized to allow a single call instead of the two different calls that were made from `new-package.jl` and `new-version.jl`.

ericphanson: ahh ok, I get it now, makes sense!

ericphanson (approved):

DilumAluthge:
Land #408 first, and then merge this and tag a new release?

docs/src/guidelines.md
DilumAluthge: FWIW, I don't think it's great for performance, but AutoMerge doesn't really have a lot of performance-sensitive code.

DilumAluthge:
bors merge

docs/src/guidelines.md
GunnarFarneback: The code isn't compiled until it's called and by that time the constants are defined (otherwise it would have been an error). It shouldn't affect the performance.

PR #408, opened 2021-05-17
Set `JULIA_PKG_PRECOMPILE_AUTO` env variable

### LINKED ISSUE #400: packages precompile twice during AutoMerge on 1.6
Since they precompile at `Pkg.add` time and we do it twice (with different depots). We should either re-use the depot or prevent precompilation at add time.

ericphanson:
Fixes #400.

I did not bump the version to avoid conflicts with #407

### DIFF
--- a/src/AutoMerge/guidelines.jl
+++ b/src/AutoMerge/guidelines.jl
@@ -661,6 +661,7 @@ function meets_version_can_be_imported(working_directory::String,
         $(pkg_add_command)
         @info("Successfully `Pkg.add`ed package");
         @info("Attempting to `import` package");
+        Pkg.precompile()
         import $(pkg);
         @info("Successfully `import`ed package");
         """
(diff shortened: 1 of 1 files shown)


ericphanson:
I want to run the integration tests to double check it only precompiles at `import` time with this PR.

bors try

ericphanson:
In https://github.com/JuliaRegistries/RegistryCI.jl/runs/2596754467?check_suite_focus=true#step:8:474 from the last merge, it says "precompiling" during the `Pkg.add` step. In https://github.com/JuliaRegistries/RegistryCI.jl/runs/2600851999#step:8:632 from this PR, I don't see any precompiling. So I think it worked!

DilumAluthge:
See the additional commit I pushed.

DilumAluthge (approved):

DilumAluthge:
bors try

DilumAluthge:
bors merge

ericphanson:
Thanks @DilumAluthge!

PR #409, opened 2021-05-18
A few tweaks to the docs for AutoMerge guidelines

DilumAluthge:
I've bumped the version number because the `stable` docs link is not updated unless we register and tag a new release.


DilumAluthge:
I've bumped the version number because the `stable` docs link is not updated unless we register and tag a new release.

ericphanson (commented):
This seems helpful! Maybe “other registries” instead of private? Since it doesn’t actually matter if they are public or private

DilumAluthge:
Done!

ericphanson (approved):

DilumAluthge:
bors merge

DilumAluthge:
bors merge

PR #423, opened 2021-06-07
Revert "Use the Blue Style guide, and automatically make PRs to fix code formatting"

DilumAluthge:
Reverts JuliaRegistries/RegistryCI.jl#414

PR #425, opened 2021-06-12
add a note about `registry_deps` with name field

johnnychen94:
It takes time to find out that this is even possible so I figure this is worth documenting.

### DIFF
--- a/docs/src/private-registries.md
+++ b/docs/src/private-registries.md
@@ -27,11 +27,12 @@ You can optionally use the registry name instead of the URL:
 - run: julia --project=.ci/ --color=yes -e 'import RegistryCI; RegistryCI.test()'
 + run: julia --project=.ci/ --color=yes -e 'import RegistryCI; RegistryCI.test(registry_deps=["General"])'
 ```
-This way Julia Pkg will download registry from Julia pkg server. If you host a private pkg server in your local network(e.g., enterprise network with firewall) and properly
-set up the environment variable `JULIA_PKG_SERVER`, then the network traffic doesn't need to pass through the proxy.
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Just FYI, I want to hold off on merging any PRs until I sort out #423

docs/src/private-registries.md
```diff
@@ -22,6 +22,17 @@ If the packages in your registry depend on packages in other registries, they sh
 + run: julia --project=.ci/ --color=yes -e 'import RegistryCI; RegistryCI.test(registry_deps=["https://github.com/JuliaRegistries/General"])'
 ```
 
+You can optionally use the registry name instead of the URL:
+```diff
+- run: julia --project=.ci/ --color=yes -e 'import RegistryCI; RegistryCI.test()'
++ run: julia --project=.ci/ --color=yes -e 'import RegistryCI; RegistryCI.test(registry_deps=["General"])'
+```
+This way Julia Pkg will download registry from Julia pkg server. If you host a private pkg server in your local network(e.g., enterprise network with firewall) and properly
```
GunnarFarneback, line 30: Technically the use of pkg server is Julia version dependent (not available at all on 1.3, available on 1.4 if `JULIA_PKG_SERVER` is set, available by default for 1.5 and 1.6 dependent on platform and exact patch version), but I'm not sure it's worth worrying about.

johnnychen94: Yeah.. adding an extra note about the pkg server version compat seems cumbersome to me. Because RegistryCI is mainly used by advanced users who already have sufficient knowledge about Pkg, I use the word "recognized" and hope they understand it :)

ericphanson (approved):

DilumAluthge (approved):

DilumAluthge:
bors merge

PR #434, opened 2021-09-16
More details regarding custom registries

SebastianM-C:
I've added some notes documenting some of the configuration details that I've recently learn about on the #github-actions slack channel.


ericphanson (approved):
These notes look hard-earned and useful to others, but they also look like something that can become out of date pretty quickly-- for example, it depends on the current configuration of the General registry, and I easily see that getting updated without this being updated. But that's already true of the docs here so I guess adding details is still a win :).

DilumAluthge:
bors merge

PR #436, opened 2021-10-05
Upper-bounded -> Upper bounded

marwahaha:
So it matches the text
> For more information, please see the "Upper bounded [compat] entries" subsection under "Additional information" below.
https://juliaregistries.github.io/RegistryCI.jl/stable/guidelines/

### DIFF
--- a/docs/src/guidelines.md
+++ b/docs/src/guidelines.md
@@ -68,7 +68,7 @@ return output_markdown
 
 ## Additional information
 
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Actually, I think that the hyphenated form ("upper-bounded") is the preferred format. So we should change all instances of "upper bounded" to "upper-bounded".

PR #440, opened 2022-01-29
Revise the commit regex.

GunnarFarneback:
This revises the commit regex to match
```
r"Commit: ([0-9a-f]+)"
```
but no longer requires that it is a markdown item on a line of its own.

The rationale is https://github.com/GunnarFarneback/LocalRegistry.jl/pull/49, which makes use of "git push options" to create a pull request (only effective on GitLab). This provides a very low complexity pull request creation mechanism but has the limitation that it cannot pass newlines.

In terms of robustness and input sanitation this change makes no difference to Registrator generated pull requests.In fact it's rather stricter since the code in total only allows the commit hash to be exactly 40 hexadecimal digits.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>", "contributors"]
-version = "7.6.0"
+version = "7.7.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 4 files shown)


ericphanson:
Sounds fine to me, but the docs need to be updated too: https://juliaregistries.github.io/RegistryCI.jl/stable/regexes/

GunnarFarneback:
Yeah, the doctests are complaining.

GunnarFarneback:
Documentation builds fine now. The Julia nightly failure is unrelated.

ericphanson (approved):

DilumAluthge:
@GunnarFarneback or @ericphanson Can one of you do the following:
1. Bump the minor version number (since this probably falls under new feature) so that we can register as soon as we merge.
2. Squash all commits into one (since Bors can't do a squash-merge).

Once the two above are done, @ericphanson please `bors merge` if this looks good to you.

ericphanson:
bors merge

PR #443, opened 2022-03-23
Allow TOML as dependency of JLL packages

giordano:
Since https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1128 we also use TOML for JLLs in some cases.

PR #444, opened 2022-03-24
Allow JLL packages to depend on TOML.

maleadt:
After https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1191, JLLs using platform augmentation will depend on TOML. Example: https://github.com/JuliaRegistries/General/pull/57214


fredrikekre:
#443 :)

PR #450, opened 2022-05-07
Bug fix: restore the "no prerelease" versions check; we never meant to remove it

DilumAluthge:
Instead of making the "no prereleases" check part of some other checks (like the "initial version" check and the "sequential version" check), this PR makes the "no prereleases" check its own top-level check.


DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
bors try

giordano (approved):
Looks good as far as I can tell, but maybe adding a test would be useful

DilumAluthge:
bors merge

DilumAluthge:
Added some tests

PR #452, opened 2022-08-02
Add the ability for registry maintainers to pass additional environment variables to the sub-processes that run `Pkg.add("Foo")` and `import Foo`

DilumAluthge:


DilumAluthge:
bors merge

PR #458, opened 2022-08-10
Registry consistency tests: the compress roundtrip check should compare `VersionRange`s (not `String`s)

### LINKED ISSUE #457: The compress roundtrip check is not 100% effective since it uses strings.
In https://github.com/JuliaRegistries/General/pull/65997 I effectively did a no-op by loading the Registry into the Pkg Registry datastructure and then saving it back out. However, this caused a big diff even though RegistryCI has a check for this https://github.com/JuliaRegistries/RegistryCI.jl/blob/b98ce77416e5e239700453f65c671c7769997c45/src/registry_testing.jl#L222-L223.

The reason for this is that the `Compress.load` is using strings and therefore sees a difference between version specs that are written differently but are in fact the same. For example in:

gnome-shell-screenshot-1s0rlw

my PR noticed that `0.2` and `0.2.0 - 0.2` where the same version range and collapsed that into the broader version range of the package above while this was missed in CI since the strings were different.

DilumAluthge:
Fixes #457

Will likely fail until https://github.com/JuliaRegistries/General/pull/65997 is merged.

### DIFF
--- a/src/registry_testing.jl
+++ b/src/registry_testing.jl
@@ -224,7 +224,7 @@ function test(path=pwd(); registry_deps::Vector{<:AbstractString}=String[])
                     compressed = RegistryTools.Compress.compress(
                         compatfile, RegistryTools.Compress.load(compatfile)
                     )
-                    mapdict = (f, dict) -> Dict(f(k, v) for (k, v) in pairs(dict))
+                    mapdict = (f, dict) -> Dict(f(k, v) for (k, v) in dict)
                     f_inner = (k, v) -> (k, Pkg.Types.VersionRange.(v))
                     f_outer = (k, dict) -> (k, mapdict(f_inner, dict))
                     Test.@test mapdict(f_outer, compressed) == mapdict(f_outer, compat)
--- a/src/registry_testing.jl
+++ b/src/registry_testing.jl
@@ -224,10 +224,10 @@ function test(path=pwd(); registry_deps::Vector{<:AbstractString}=String[])
                     compressed = RegistryTools.Compress.compress(
                         compatfile, RegistryTools.Compress.load(compatfile)
                     )
-                    mapdict = (f, dict) -> Dict(f(k, v) for (k, v) in dict)
-                    f_inner = (k, v) -> (k, Pkg.Types.VersionRange.(v))
-                    f_outer = (k, dict) -> (k, mapdict(f_inner, dict))
-                    Test.@test mapdict(f_outer, compressed) == mapdict(f_outer, compat)
+                    mapvalues = (f, dict) -> Dict(k => f(v) for (k, v) in dict)
+                    f_inner = v -> Pkg.Types.VersionRange.(v)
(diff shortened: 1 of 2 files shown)


DilumAluthge:
There's surely a simpler way to implement this. Is there a way to apply `map` over just the values of a Dict?

src/registry_testing.jl
```diff
@@ -218,11 +218,16 @@ function test(path=pwd(); registry_deps::Vector{<:AbstractString}=String[])
...
                     Test.@test load_compat(compatfile, vnums)
-                    # Make sure the content roundtrips through decompression/compression
+                    # Make sure the content roundtrips through decompression/compression.
+                    # However, before we check for equality, we change the compat ranges
+                    # from `String`s to `VersionRanges`.
                     compressed = RegistryTools.Compress.compress(
                         compatfile, RegistryTools.Compress.load(compatfile)
                     )
-                    Test.@test compressed == compat
+                    mapdict = (f, dict) -> Dict(f(k, v) for (k, v) in pairs(dict))
```
GunnarFarneback, line 227: `pairs` doesn't do anything here, it's fine to iterate directly over the dict.

GunnarFarneback:
If it's fine to mutate you can use the method:
```
  map!(f, values(dict::AbstractDict))

  Modifies dict by transforming each value from val to f(val). Note that the type of dict
  cannot be changed: if f(val) is not an instance of the value type of dict then it will be
  converted to the value type if possible and otherwise raise an error.
```
I'm not sure it will simplify much though.

src/registry_testing.jl
```diff
@@ -218,11 +218,16 @@ function test(path=pwd(); registry_deps::Vector{<:AbstractString}=String[])
...
                     Test.@test load_compat(compatfile, vnums)
-                    # Make sure the content roundtrips through decompression/compression
+                    # Make sure the content roundtrips through decompression/compression.
+                    # However, before we check for equality, we change the compat ranges
+                    # from `String`s to `VersionRanges`.
                     compressed = RegistryTools.Compress.compress(
                         compatfile, RegistryTools.Compress.load(compatfile)
                     )
-                    Test.@test compressed == compat
+                    mapdict = (f, dict) -> Dict(f(k, v) for (k, v) in dict)
```
GunnarFarneback, line 227: There's no need, nor intention, to ever transform the keys, so
```
(f, dict) -> Dict(k => f(v) for (k, v) in dict)
```
seems more natural to me. (And corresponding changes of the mapping functions.)

GunnarFarneback (approved):

DilumAluthge:
bors merge

KristofferC:
Did this fail on the registry prior to the normalisation PR?

DilumAluthge:
It did not.

KristofferC:
Okay, I don't think it is 100% effective then since there clearly is a way to compress it further. I'll try see what the difference is (in a couple of days).

PR #461, opened 2022-08-14
Unit tests: whenever we set a temp `JULIA_DEPOT_PATH`, we should also unset `JULIA_PROJECT`

DilumAluthge:


DilumAluthge:
bors merge

PR #467, opened 2022-08-14
CI: use a global concurrency group to make sure that only one integration test job is running at a time

DilumAluthge:


DilumAluthge:
bors merge

PR #484, opened 2022-11-07
add `is_pkgname_similar` 

anandijain:
i refactored a bit to make it easier to check if your package name is too similar to existing package names

i think tests rely on having the uncompressed General, so I don't know if tests pass yet. 

the reason I wanted this is so I could build a `pkg_search` function that lets me search for packages from the repl

very sorry for juliaformatter


ericphanson (commented):
can you undo the unrelated formatting changes? it makes it harder to review

PR #492, opened 2023-01-24
Check that there are "enough" tests and documentation

MarkNahabedian:
These changes introduce the new Guideline
guideline_linecounts_meet_thresholds that uses PackageAnalyzer to
collect line counts from the candidate package and compare against
specified thresholds.  The thresholds are provided ad these keyword
arguments to `AutoMerge.run`:

 * src_min_lines:       Minimum number of lines of source code
 * readme_min_lines:  % Minimum number of lines in the README file
 * test_min_lines:    % Minimum number of lines of code in the test directory
 * doc_min_lines:     # Minimum number of lines of documentation

Those marked with a % can be expressed as a count, or as a fraction
of the number of lines of source code.  For test_min_lines and
doc_min_lines, the denominator of the fraction also includes the
number of lines of test code.

Additional work:

All Guidelines are collected into `ALL_GUIDELINES` to make them easier
to identify and count.

`run` creates a Dict with which to opaquely pass parameters for the
various Guidelines down through `pull_request_build` and `check!`.
That Dict parameter has beed added to all Guideline `check` functions,
most of which ignore it for now.

Report which keyword arguments are missing from GitHubAutoMergeData
using @warn.

The depot setup code from the AutoMerge.parse_registry_pkg_info
testset has been extracted to the new function `setup_depot` so that
it can be used in other tests.


MarkNahabedian:
How should I deal with the PackageAnalyzer version issue? PackageAnalyzer has a constraint that Julia version be 1.6 or greater.

ericphanson:
Could we drop support for pre-Julia 1.6 in RegistryCI @DilumAluthge?

DilumAluthge:
I really would prefer that we keep support for all Julia versions >= Julia 1.3. That way, we can add new registry consistency checks in the future, and those checks will be run on Julia 1.3.

PackageAnalyzer is only needed for AutoMerge, right? Can you register an empty version of PackageAnalyzer that is compatible with `julia = "1"`? It would contain no source code.

Alternatively, can we make PackageAnalyzer a weak dep for Julia 1.9+? Then, it will be ignored for Julia versions prior to 1.9, right?

src/AutoMerge/guidelines.jl
```diff
@@ -857,6 +866,65 @@ function meets_version_can_be_imported(
...
+    src_min_lines    = get(guideline_parameters, :src_min_lines,   10)
+    readme_min_lines = get(guideline_parameters, :readme_min_lines, 5)
+    test_min_lines   = get(guideline_parameters, :test_min_lines,   0.1f0)
+    doc_min_lines    = get(guideline_parameters, :doc_min_lines,    0.1f0)
+    issues = []
+    src_line_count = PackageAnalyzer.count_julia_loc(analysis, "src")
+    if !meets_threshold(readme_min_lines,
+                         PackageAnalyzer.count_readme(analysis),
+                         src_line_count)
+        push!(issues, "Package README file is too short")
```
ericphanson, line 904: I think we should make these errors include the counts, so they are a bit more informative:
```suggestion
        push!(issues, "Package README file is too short. Counted $(src_line_count) lines, but at least $(readme_min_lines) lines are required.")
```

src/AutoMerge/public.jl
```diff
@@ -61,6 +61,11 @@ RegistryCI.AutoMerge.run(
     additional_check_runs = String[],
     check_license = true,
     public_registries = String["https://github.com/HolyLab/HolyLabRegistry"],
+    # Parameters for guideline_linecounts_meet_thresholds
+    src_min_lines::Integer,
+    readme_min_linges::Union{Integer, AbstractFloat},
+    test_min_lines::Union{Integer, AbstractFloat},
+    doc_min_lines::Union{Integer, AbstractFloat},
```
ericphanson, line 68: I think we should put the default values here

src/AutoMerge/types.jl
```diff
@@ -100,18 +100,30 @@ struct GitHubAutoMergeData
...
+    mismatch = setdiff(Set(keys(kwargs)), Set(fields))
+    if !isempty(mismatch)
+        @warn "Keyword/field mismatch while constructing a GitHubAutoMergeData: $(join(mismatchmissing, ", "))"
+    end
+    always_assert(isempty(mismatch))
     always_assert(kwargs[:authorization] ∈ (:normal, :jll))
     return GitHubAutoMergeData(getindex.(Ref(kwargs), fields)...)
 end
 
+ALL_GUIDELINES = []
```
ericphanson, line 125: Do we need this?

src/utils.jl
```diff
@@ -25,3 +27,24 @@ function with_temp_depot(f::Function)
...
+    if v == nothing
+        return default
+    end
+    m = match(r"(?<count>^[0-9]+)$", v)
+    if m != nothing
+        return parse(Int64, m["count"])
+    end
+    @warn "Value $v of environment variable $envvar is not a line count.  Using default of $default"
+    default
+end
```
ericphanson, line 49: I think we don't need this, since we aren't using ENV variables for these

src/AutoMerge/guidelines.jl
```diff
@@ -857,6 +866,65 @@ function meets_version_can_be_imported(
...
+    analysis = PackageAnalyzer.analyze(pkg_code_path;)
+    # @info analysis
+    if isempty(analysis.lines_of_code)
+        return false, "PackageAnalyzer didn't find any lines of code."
+    end
+    # get parameters:
+    src_min_lines    = get(guideline_parameters, :src_min_lines,   10)
+    readme_min_lines = get(guideline_parameters, :readme_min_lines, 5)
+    test_min_lines   = get(guideline_parameters, :test_min_lines,   0.1f0)
+    doc_min_lines    = get(guideline_parameters, :doc_min_lines,    0.1f0)
```
ericphanson, line 898: I think to start we should make these all fixed values, not fractions. Maybe:
```suggestion
    test_min_lines   = get(guideline_parameters, :test_min_lines,   10)
    doc_min_lines    = get(guideline_parameters, :doc_min_lines,    10)
```
I also think we shouldn't require docs & readme, since some packages document their code via the readme. I think we should ask for 10-20 lines of docs OR readme

src/AutoMerge/util.jl
```diff
@@ -292,3 +292,34 @@ function get_all_non_jll_package_names(registry_dir::AbstractString)
...
+
+function meets_threshold(threshold::Integer, numerator::Integer, ::Integer)
+    # @info("meets_threshold", threshold, numerator, numerator >= threshold)
+    numerator >= threshold
+end
+
+function meets_threshold(threshold::AbstractFloat, numerator::Integer, denominator::Integer)
+    # @info("meets_threshold", threshold, numerator, denominator, (numerator / denominator) >= threshold)
+    (numerator / denominator) >= threshold
+end
```
ericphanson, line 324: I'm a bit worried about this; I think it's not a good idea to have totally different semantics for integers vs floats, since e.g. `10.0 == 10` but here they would do totally different things. I would suggest starting with only fixed values, and we could add fractions later if we need it. I think just fixed values would likely be enough though.

ericphanson (commented):
thank you for your work on this! I will try to get PackageAnalyzer working on 1.3. In the meantime I've left some comments on other parts of the PR.

src/AutoMerge/types.jl
MarkNahabedian: I found it handy to be able to lust and count all of the guidelines.

I don't see that it's doing any harm.

I'm about to submit a cleanup issue where this will play a significant role.

src/utils.jl
MarkNahabedian: I don't know what you're looking at, but I think this is already gone.  I'll make sure.  I'm prettuly sure I left no changes in src/utils, but moved the remaining generic functions to src/AutoMerge.util.

src/AutoMerge/util.jl
MarkNahabedian: What if I assert that the fliats are between 0 and 1 inclusive?

I feel pretty strongly about fraction support.  10,000 lines of code should not get away with ten lines of documentation or tests.

DilumAluthge: Let's just start with absolute (non-fraction) thresholds, and see how that goes first.

src/utils.jl
MarkNahabedian: I meant to emove it, but apparently didn't.  Gone now.

src/AutoMerge/util.jl
MarkNahabedian: Assertions on thresholds are in place.

src/AutoMerge/guidelines.jl
```diff
@@ -25,7 +26,8 @@ const guideline_compat_for_julia = Guideline(;
         "There is an upper-bounded `[compat]` entry for `julia` that ",
         "only includes a finite number of breaking releases of Julia.",
     ),
-    check=data -> meets_compat_for_julia(data.registry_head, data.pkg, data.version),
+    check=(data, guideline_parameters) ->
```
GunnarFarneback, line 29: Why can't `guideline_parameters` be included in `data`?
Or well, it looks like there is new fields for them in `data`, so why is it also an extra argument for the checks?

GunnarFarneback:
> I also think we shouldn't require docs & readme, since some packages document their code via the readme. I think we should ask for 10-20 lines of docs OR readme

Agreed, it's rather pointless to require manual intervention for packages with README-only documentation or a very thin README pointing to the docs.

Project.toml
```diff
@@ -29,6 +31,7 @@ GitHub = "5.2"
 HTTP = "0.8, 0.9.1, 1"
 JSON = "0.19, 0.20, 0.21"
 LicenseCheck = "0.2"
+PackageAnalyzer = "1"
```
ericphanson, line 34: This will let us install a dummy version of the package (registered at v0.2) on Julia 1.3 to 1.5:
```suggestion
PackageAnalyzer = "0.2, 1"
```
xref https://github.com/JuliaRegistries/General/pull/76413

src/AutoMerge/types.jl
MarkNahabedian: Here's that new issue:
https://github.com/JuliaRegistries/RegistryCI.jl/issues/493

test/automerge-unit.jl
```diff
@@ -616,4 +630,25 @@ end
         result = has_osi_license_in_depot("VisualStringDistances")
         @test !result[1]
     end
+
+    @testset "AutoMerge.linecounts_meet_thresholds" begin
```
ericphanson, line 634: I think we need to ensure this test only runs on 1.6+ with an `if VERSION >= v"1.6"` check

src/AutoMerge/guidelines.jl
MarkNahabedian: I meant to remove them from GitHubAutoMergeData once I decided it wasn't appropriate for that struct to accrete non-GutHub API related stuff.

GitHubAutoMergeData isn't created untill pull_requedt_build so guideline parameters would need to be passed to pull_reqyest_build.

Sorry if there are typos or misremembered identifiers.  I'm typing on my phone on a very bouncy bus ride.

MarkNahabedian: Linecount parameters have been removed from GitHubAutoMergeData.

test/automerge-unit.jl
MarkNahabedian: Done.

MarkNahabedian:
Concerning requireing both docs and readme, I'm not able to respond in-line.

I agree that Either is sufficient and they're not both necessary.  Docs isn't really sufficiholds.ent unless there's a link from the readme.    I don't think PackageAnalyzer checks for that, and I'm willing to let the doc link slide for now.

Does someone want to propose logic for the docs/readme check?

Right now they must both exceed their respective thresholds?

Should it be sufficient that either meets its threshold?

I was originally thinking that the readme threshold would be a lower bar -- readme being a brief explanation while documentation provides full details about how to use the package.  I feel that one is being inconsiderate of ones users if they don't provide good documentation.

Maybe if the docs threshold isn't met then the readme linecount should meet the docs threshold instead of the readme threshold?

MarkNahabedian:
While we're critiqueing thresholds, I'll not that lines of test code is a lame surrogate for test coverage.  Maybe that will be the topic of a future guideline, whence we can change the tests threshold for this guideline to 0.

src/AutoMerge/guidelines.jl
GunnarFarneback: Most of the fields in the struct are not GitHub specific. Once someone implements support for some more backend, there are good reasons to refactorize the struct, but the real point of the struct is to collect the data needed by the guidelines so we don't have to pass around lots of arguments.

MarkNahabedian: There is no reason to not start that refactoring now.

The real point of the struct, if the name is to be believed, is to communicate with the GitHub API and provide it's results.

That it has been co-opted as a repository for more general parameters is unfortunate.  I was able to restrain myself from doing that further.

Like I said before, it is created too late to avoid having to declare arguments for each guideline parameter for pull_request_build.  I'm willing to migrate parameters from function arguments and GirHubAutoMergeData to the fict, but not in this PR.

Most of the work I've put in to this PR was in tracing how existing guideline parameters fliwed through the code and tracking how they might get modified during that journey.  I hope to make that much clearer for the next guy.

DilumAluthge:
I'd like to do refactoring in separate PRs from feature additions.

MarkNahabedian:
https://github.com/MarkNahabedian/AnalyzeRegistryPackages.jl provides an analysis of the packages in the general registry.

The scatter plots should probably have excluded jlls, but they don't yet.

src/AutoMerge/util.jl
MarkNahabedian: I introduced three new parameters for specifying ratios.  The _min_lines parameters just check absolute line count.  The new _fraction parameters check the ratio of the respective line count to number of source lines or sum of source lines and the respective numerator.

MarkNahabedian:
I don't think there is anything more I can do here until there's a discussion about how to deasl with the docs v. README dichotomy mentioned above.

Comments?

MarkNahabedian:
After a lot of muddle headed thinking, I have some plots that might help to inform what values we might choose for the various thresholds.  The code to collect the data and the plots are at https://github.com/MarkNahabedian/AnalyzeRegistryPackages.jl.

There's also a TSV file of registered repositories that were unreachable in two attempts.

MarkNahabedian:
Ping.

Is anyone going to review this?

If there is something more that I should do, please inform me.

DilumAluthge:
@ericphanson Would you be able to review this?

DilumAluthge:
@MarkNahabedian Can you rebase on the latest master, and fix the merge conflicts?

MarkNahabedian:
@DilumAluthge:. I've never used `rebase` before, and from what I've read, I'm not comfortable with it.

I saw the merge conflict and managed to resolve some of it.  I don't see how to resolve the one remaining diff though.

src/AutoMerge/public.jl
```diff
@@ -102,7 +107,26 @@ function run(;
     public_registries::Vector{<:AbstractString}=String[],
     read_only::Bool=false,
     environment_variables_to_pass::Vector{<:AbstractString}=String[],
+    # Four parameters for guideline_linecounts_meet_thresholds
```
ericphanson, line 110:
```suggestion
    # Parameters for guideline_linecounts_meet_thresholds
```

src/AutoMerge/guidelines.jl
```diff
@@ -860,6 +869,77 @@ function meets_version_can_be_imported(
...
+ * src_min_lines:       Minimum number of lines of source code
+ * readme_min_lines:    Minimum number of lines in the README file
+ * test_min_lines:      Minimum number of lines of code in the test directory
+ * doc_min_lines:       Minimum number of lines of documentation
+ * readme_min_fraction  minimum ratio of readme lines to src lines
+ * test_min_fraction    minimum ratio of test lines to src lines
+ * doc_min_fraction     minimum ratio of doc to src lines
+
+For test_min_fraction and doc_min_fraction, the denominator of the fraction
+also includes the number of lines of test or docs respectively.
```
ericphanson, line 885:
```suggestion

 * `src_min_lines`:       Minimum number of lines of source code
 * `readme_min_lines`:    Minimum number of lines in the README file
 * `test_min_lines`:      Minimum number of lines of code in the test directory
 * `doc_min_lines`:       Minimum number of lines of documentation
 * `readme_min_fraction`:  minimum ratio of readme lines to src lines
 * `test_min_fraction`:    minimum ratio of test lines to src lines
 * `doc_min_fraction`:     minimum ratio of doc to src lines

For `test_min_fraction` and `doc_min_fraction`, the denominator of the fraction
also includes the number of lines of test or docs respectively.
```

src/utils.jl
```diff
@@ -1,3 +1,5 @@
+using Logging
+
```
ericphanson, line 2: why is this needed?

test/automerge-unit.jl
```diff
@@ -624,4 +640,66 @@ end
...
+            tmp_depot = setup_depot()
+            # I don't know of a package we can rely on as static.  Test
+            # this package and update the thresholds and issue messages as
+            # needed.
+            package_name = "RegistryCI"
+            pkg_code_path = pkgdir_from_depot(tmp_depot, package_name)
+            # The PackageAnalyzer results last measured are
+            # src: 3412, tests: 1336, docs: 309, readme: 23
+            guideline_parameters = Dict{Symbol, Any}(
+                # Initial setup for all metrrics to pass:
```
ericphanson, line 656:
```suggestion
                # Initial setup for all metrics to pass:
```

ericphanson (commented):
I think https://github.com/JuliaRegistries/RegistryCI.jl/pull/492/files#r1087494519 should still be addressed. If we want to rename the struct in a separate PR that seems fine, but this PR (adding a specific guideline) should not also be restructuring the data flow through all the guidelines!

Otherwise, the implementation w/ separate parameters for counts & fractions looks good to me. Gives the most choice to the downstream registry.

DilumAluthge:
> I think https://github.com/JuliaRegistries/RegistryCI.jl/pull/492/files#r1087494519 should still be addressed. If we want to rename the struct in a separate PR that seems fine, but this PR (adding a specific guideline) should not also be restructuring the data flow through all the guidelines!

I agree. In general, refactor/restructure PRs should be separate from "add new guideline" PRs.

@MarkNahabedian If you like, you can first make a PR to do the desired refactoring, and then once that PR has been reviewed and merged, you can rebase this PR on master. So that'll let you use the benefits of the refactoring in this PR.

src/AutoMerge/public.jl
MarkNahabedian: I'll update the count to "Seven".

I want to avoid confusion if additional parameters are added without a comment.

src/utils.jl
MarkNahabedian: I might have needed it for debugging at some point.

I see Logging being used elsewhere, but I don't see it needed here.  I'll take it out.

MarkNahabedian:
> > I think https://github.com/JuliaRegistries/RegistryCI.jl/pull/492/files#r1087494519 should still be addressed. If we want to rename the struct in a separate PR that seems fine, but this PR (adding a specific guideline) should not also be restructuring the data flow through all the guidelines!
> 
> I agree. In general, refactor/restructure PRs should be separate from "add new guideline" PRs.
> 
> @MarkNahabedian If you like, you can first make a PR to do the desired refactoring, and then once that PR has been reviewed and merged, you can rebase this PR on master. So that'll let you use the benefits of the refactoring in this PR.

I don't believe I've done any refactoring here.  I have proposed refactoring in an issue cited above.

The codebase shows that how parameters are passed from the entry point to a guideline is arbitrary.  Yes, I added a third way for parameters to be passed, but at some point in the past someone added a second.

I've spent more time on this change than I expected too.  Removing passing a parameter dict feels like a step backwards.  At the risk of seeming uncooperative, I don't want to spend any of my time on it.  Any of the package maintainers can do so if they actually think this is important.

MarkNahabedian:
I tried createing another fork for the guideline parameter change, but GitHub wouldn't let me.

I don't know how to proceed.

MarkNahabedian:
A small part of this PR is now broken out to https://github.com/JuliaRegistries/RegistryCI.jl/pull/506.

Please review.

MarkNahabedian:
Part of this change has been merged as https://github.com/JuliaRegistries/RegistryCI.jl/pull/506

PR #495, opened 2023-02-17
Allow the `Deps.toml`, `WeakDeps.toml`, and `Compat.toml`, and `WeakCompat.toml` files to not exist

### LINKED ISSUE #494: LoadError: "/home/runner/work/General/General/I/InstanceOfSingletonType/Deps.toml": No such file
https://github.com/JuliaRegistries/General/actions/runs/4028452961/jobs/6925352444#step:16:69

I guess this is caused by the fact that my package has no dependencies?

DilumAluthge:
Fixes https://github.com/JuliaRegistries/RegistryTools.jl/issues/81
Fixes #494

### DIFF
--- a/src/AutoMerge/guidelines.jl
+++ b/src/AutoMerge/guidelines.jl
@@ -32,7 +32,7 @@ function meets_compat_for_julia(working_directory::AbstractString, pkg, version)
     package_relpath = get_package_relpath_in_registry(;
         package_name=pkg, registry_path=working_directory
     )
-    compat = Pkg.TOML.parsefile(joinpath(working_directory, package_relpath, "Compat.toml"))
+    compat = maybe_parse_toml(joinpath(working_directory, package_relpath), "Compat.toml")
     # Go through all the compat entries looking for the julia compat
     # of the new version. When found, test
     # 1. that it is a bounded range,
@@ -80,8 +80,8 @@ function meets_compat_for_all_deps(working_directory::AbstractString, pkg, versi
     package_relpath = get_package_relpath_in_registry(;
         package_name=pkg, registry_path=working_directory
     )
-    deps = Pkg.TOML.parsefile(joinpath(working_directory, package_relpath, "Deps.toml"))
-    compat = Pkg.TOML.parsefile(joinpath(working_directory, package_relpath, "Compat.toml"))
+    deps = maybe_parse_toml(joinpath(working_directory, package_relpath), "Deps.toml")
+    compat = maybe_parse_toml(joinpath(working_directory, package_relpath), "Compat.toml")
     # First, we construct a Dict in which the keys are the package's
     # dependencies, and the value is always false.
     dep_has_compat_with_upper_bound = Dict{String,Bool}()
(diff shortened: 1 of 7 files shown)


DilumAluthge:
bors try

DilumAluthge:
bors try-

DilumAluthge:
bors try

sjkelly:
Thanks @DilumAluthge !

DilumAluthge:
Bump @KristofferC

src/AutoMerge/toml.jl
```diff
@@ -0,0 +1,18 @@
+function maybe_parse_toml(directory::AbstractString, file::AbstractString)
```
KristofferC, line 1: I would just write this as

```julia
maybe_parse_toml(f) = isfile(f) ? TOML.parsefile(f) : Dict{String, Any}()
```

All these error checking etc seems like it should be up to the caller. And most functions take a full path to a file, not a `directory` + `file`.

DilumAluthge: Done:
1. `maybe_parse_toml` takes the full path to a file.
2. `maybe_parse_toml` is defined in one line, using a ternary.
3. The error checking is done by each caller.

src/AutoMerge/toml.jl
```diff
@@ -0,0 +1,18 @@
...
+        "Deps.toml",
+        "WeakDeps.toml",
+    )
+    if !(file in allowed_filenames)
+        msg = "Filename is not in the allowed list: $(file)"
+        throw(ErrorException(msg))
+    end
+    full_path = joinpath(directory, file)
+    if !ispath(full_path)
+        @warn "TOML file does not exist; returning an empty dict" directory file full_path
```
KristofferC: And why is this a warning? Getting an empty dict when the file does not exist is the point of the function and why you call it in the first place. So why should it warn?

DilumAluthge: The warning is now removed.

src/AutoMerge/guidelines.jl
```diff
@@ -32,7 +32,9 @@ function meets_compat_for_julia(working_directory::AbstractString, pkg, version)
     package_relpath = get_package_relpath_in_registry(;
         package_name=pkg, registry_path=working_directory
     )
-    compat = Pkg.TOML.parsefile(joinpath(working_directory, package_relpath, "Compat.toml"))
+    compat_file = joinpath(working_directory, package_relpath, "Compat.toml")
+    assert_allowed_to_not_exist(compat_file)
```
KristofferC, line 36: I don't really understand this `assert`. It will never fire because you are calling it with a compat file just defined above. The act of calling `maybe_parse_toml` means that the file is allowed to not exist. So I am not sure the `assert` adds anything?

DilumAluthge: I've removed the asserts.

DilumAluthge:
Bump @KristofferC

fredrikekre (approved):

DilumAluthge:
bors merge

DilumAluthge:
@KristofferC If you have a chance, it would be great if you could do one final review (even if it's post-merge), and I can make the requested changes in a follow-up PR.

I'm just merging this now so we can get some kind of fix deployed, because there are PRs stalled out on General.

PR #498, opened 2023-02-25
Point to Slack `#pkg-registration` for manual merges

### LINKED ISSUE #491: Update manual merge guideline
The current guideline:
https://github.com/JuliaRegistries/RegistryCI.jl/blob/b9018d8efbc0dd8d1504c294cd71ce5bc677057d/src/AutoMerge/util.jl#L102-L103
In my experience, posting a comment often doesn't help: nobody notices it ever.

Maybe, add "... and post the PR link to Slack #pkg-registration channel" to this message? Or something else, whatever is the preference of `General` maintainers.

goerz:
As [discussed on Slack earlier today](
https://julialang.slack.com/archives/C6M4DQA5P/p1677337269883019?thread_ts=1677284631.873199&cid=C6M4DQA5P), and in issue #491, manual merges of new packages will usually only happen after someone pings the `#pkg-registration` channel on Slack. For new package developers who may not be aware of the Julia Slack, this may cause considerable frustration, potentially alienating them from the Julia ecosystem. Until there is some other workflow in place to handle this entirely within GitHub, it seems best to explicitly recommend that people post a message on Slack to get help with the manual merge.

This PR amends the instructions in the "If you do not want to fix the AutoMerge issues..." bot message, asking the author to post to `#pgk-registration` in addition to commenting on the GitHub issue.

Closes #491


DilumAluthge:
@goerz Since this is a new feature, can you also bump the minor version number in the `Project.toml` file?

ericphanson:
I think this message makes sense for General but not other registries using RegistryCI. Can we make it optional? We’d need to thread through a keyword argument from the public api functions.

I also think we should default to false and then opt-in in the workflow in General or else tag it as breaking, since private registries will want to opt out.

goerz:
Ah, yes, I was assuming this was only used for the General registry. Making it optional totally makes sense then, but I might need some guidance on how to do that.

goerz:
Ok, I added a commit that makes the message optional (and off by default), but I'm not sure what to do to switch it on for the General registry. I'm not quite sure where the `comment_text_fail` function is called from or how exactly it ties into the bot for General.

src/AutoMerge/util.jl
```diff
@@ -84,7 +84,7 @@ function parse_registry_pkg_info(registry_path, pkg, version=nothing)
     return (; uuid=uuid, repo=repo, subdir=subdir, tree_hash=tree_hash)
 end
 
-function _comment_disclaimer()
+function _comment_disclaimer(point_to_slack::Bool=false)
```
DilumAluthge, line 87: It would probably be best to make this a keyword argument.

goerz: Sure! Do we want to show the message both for new packages and new versions? Right now it's only for new versions, but I can easily add the flag for the `comment_text_fail(::NewVersion, ...` method as well

DilumAluthge: Let's do it for both.

goerz: Do you want it to be a keyword one level up, in `comment_text_fail`, as well? That function already has `suggest_onepointzero` as a positional argument, which seems similar. But adding `point_to_slack` replaces `version` as the fourth positional argument there. So, make `point_to_slack` a keyword argument after `version`?

goerz: Okay, I made it a keyword argument throughout in the latest commit https://github.com/JuliaRegistries/RegistryCI.jl/pull/498/commits/4d5591ca9066bc1c41d23a43ad9b96e132cc2faa, but I can roll that back or change it if you have something different in mind.

DilumAluthge: Yeah, in hindsight, `suggest_onepointzero` should have also been a kwarg.

goerz:
It looks like the syntax I'm using for the keyword argument fails in Julia < 1.5. I'm assuming I have to change `_comment_disclaimer(; point_to_slack)` to `_comment_disclaimer(; point_to_slack=point_to_slack)`?

src/AutoMerge/util.jl
```diff
@@ -184,7 +191,7 @@ function comment_text_fail(
         _please_read_these_documents,
         "The following guidelines were not met:\n\n",
         reasons_formatted,
-        _comment_disclaimer(),
+        _comment_disclaimer(; point_to_slack),
```
DilumAluthge, line 194: We need to keep Julia 1.3 support.

```suggestion
        _comment_disclaimer(; point_to_slack=point_to_slack),
```

goerz: Done, let's see if that runs through…

DilumAluthge:
This is the public entrypoint to the package: https://github.com/JuliaRegistries/RegistryCI.jl/blob/74c541850b665654864cb2c9fae455b94ef4e385/src/AutoMerge/public.jl#L67

The `suggest_onepointzero` variable is a good example here. You can watch how it starts at the top-level (https://github.com/JuliaRegistries/RegistryCI.jl/blob/74c541850b665654864cb2c9fae455b94ef4e385/src/AutoMerge/public.jl#L90) and then it is propagated down until it reaches the place where it is used.

DilumAluthge:
Looks like there's a merge conflict that needs to be resolved. Probably the `version` line in `Project.toml.

goerz:
> This is the public entrypoint to the package:
> 
> https://github.com/JuliaRegistries/RegistryCI.jl/blob/74c541850b665654864cb2c9fae455b94ef4e385/src/AutoMerge/public.jl#L67
> 
> The `suggest_onepointzero` variable is a good example here. You can watch how it starts at the top-level [...] and then it is propagated down until it reaches the place where it is used.

Got it. The latest commit now exposes `point_to_slack` in the `Automerge.run` function. If I understand correctly, once this PR is merged, I would have to follow up with a PR to the General registry to add `point_to_slack = true` to the [`automerge.yml`](https://github.com/JuliaRegistries/General/blob/905f69b5574bbde6c0011757f4c44f468ba449dc/.github/workflows/automerge.yml#L80).

ericphanson:
bors try

src/AutoMerge/public.jl
```diff
@@ -27,6 +27,7 @@ Run the `RegistryCI.AutoMerge` service.
 - `master_branch`: name of `master_branch`, e.g you may want to specify this to `"main"` for new GitHub repositories.
 - `master_branch_is_default_branch`: if `master_branch` specified above is the default branch.
 - `suggest_onepointzero`: should the AutoMerge comment include a suggestion to tag a 1.0 release for v0.x.y packages.
+- `point_to_slack`: should the AutoMerge comment recommend sending a message to the `#pgk-registration` Julia-Slack channel when auto-merging is not possible.
```
ericphanson, line 30:
```suggestion
- `point_to_slack`: should the AutoMerge comment recommend sending a message to the `#pkg-registration` Julia-Slack channel when auto-merging is not possible.
```

ericphanson (approved):
Lgtm, besides the small typo.

Yeah, to get this deployed, we first merge this+register this to get a new RegistryCI release, then update the manifests used by General (there’s a workflow for that) so it uses the new release, and then update the workflow there to use the new keyword argument.


However we also don’t have an integration test here that actually uses the new functionality. I think we should either add one before merging, or else we should first deploy to our “staging” workflow, https://github.com/JuliaRegistries/General/blob/master/.github/workflows/automerge_staging.yml.

So that would be:
1. Merge + release RegistryCI 
2. Update RegistryCI version used in General
3. Update staging workflow to have new comment text (add kwarg)
4. General maintainer turns on staging workflow
5. Try that for a bit
6. general maintainer turns off staging workflow
7. Update main workflow to add comment text kwarg

goerz:
I can look into adding an integration test. I'll see if I can figure out `test/automerge-integration.jl`, but if it's not obvious, I might need some guidance…

ericphanson:
I think it might just be a matter of adding `point_to_slack` as another option here https://github.com/JuliaRegistries/RegistryCI.jl/blob/388a266e31f7e380db9ac4f3d972beabe08b91ba/test/automerge-integration.jl#L37 and then setting it to true for some of the following variants and setting it to false for some. I don’t think we need a new variant for this.

goerz:
Ok, I've added it to the integration test

ericphanson:
I think it still needs to be passed in below, like here https://github.com/JuliaRegistries/RegistryCI.jl/pull/498/files#diff-e3199c147438f7c1bdb5c32ba3730b317681ccc3f777bf1f4881f766c66400b7R226

ericphanson:
Bors try

goerz:
Duh… sorry about that! Should be fixed now. I've added the `point_to_slack` only to the first `AutoMerge.run` (in the [location you were pointing out](https://github.com/JuliaRegistries/RegistryCI.jl/pull/498/files#diff-e3199c147438f7c1bdb5c32ba3730b317681ccc3f777bf1f4881f766c66400b7R226)).

I'm not quite sure what the second `withenv` farther below does (with the two calls to `AutoMerge.run`), but it doesn't have `check_license` either, nor does it deal with pass/fail; so I left it alone. It might be worth adding some comments to explain the intent of that test, but that has nothing to do with this PR.

goerz:
One more quick fix to update the `@info "Performing integration tests with settings" …` line. Is the `Bors try` something I can/should trigger, or is that only available to maintainers?

ericphanson:
Only available to maintainers for security reasons but I can trigger it again. Need to wait for it to finish though otherwise it seems to get confused. Are you happy with it now? Next time I can ask bors to merge which will run integration tests then merge if everything passes.

goerz:
Yeah, I think it should be ready now

ericphanson:
Ok great, thanks for the contribution!

bors merge

goerz:
Thanks @DilumAluthge and @ericphanson for your help on this!

So should I do anything to follow up on this, after the `RegistryCI v8.2.0` release goes through? Should I start a PR to the General registry sometime this week to update [`automerge_staging.yml`](https://github.com/JuliaRegistries/General/blob/master/.github/workflows/automerge_staging.yml), or [`automerge.yml`](https://github.com/JuliaRegistries/General/blob/master/.github/workflows/automerge.yml) directly?

PR #499, opened 2023-02-25
Fix breakage in the `julia_compat` function

DilumAluthge:
Introduced in #495 

Ref https://github.com/JuliaRegistries/General/pull/78512#issue-1599785208


DilumAluthge:
Okay, so at least I have a unit test now that reproduces the error.

DilumAluthge:
bors merge

PR #500, opened 2023-03-09
Require at least RegistryTools 2.2, and fix the integration tests for the new `jll/` directory

DilumAluthge:
So that we get the new location of `_jll` packages.


DilumAluthge:
bors merge

PR #506, opened 2023-04-28
setup_depot for setting up a temporary depot for testing.

MarkNahabedian:
I was asked to break up https://github.com/JuliaRegistries/RegistryCI.jl/pull/492 into several parts.

This PR introduces the function `setup_depot` which pulls out the code for creating a temporary depot during testing so that it is easier to use in other places.


test/automerge-unit.jl
```diff
@@ -10,6 +10,28 @@ using TimeZones
 
 const AutoMerge = RegistryCI.AutoMerge
 
+
+TEMP_DEPOT_FOR_TESTING = nothing
+
+function setup_depot()::String
```
ericphanson, line 16:
```suggestion
function setup_global_depot()::String
```

ericphanson (commented):
I would name it `setup_global_depot` so it's clear that 1 depot is shared by all invocations of the function (rather than a separate one each time). Otherwise, looks good to me!

test/automerge-unit.jl
```diff
@@ -581,20 +603,12 @@ end
     @testset "`AutoMerge.meets_version_has_osi_license`" begin
         # Let's install a fresh depot in a temporary directory
         # and add some packages to inspect.
-        tmp_depot = mktempdir()
+        tmp_depot = setup_depot()
```
ericphanson, line 606:
```suggestion
        tmp_depot = setup_global_depot()
```

ericphanson:
Bors merge

PR #507, opened 2023-04-30
Pass guiedline parameters in a dict

MarkNahabedian:
This change simplifies the passing of paramneters from the various `Guideline`s from `run` down to the various `Guideline.check`functions.

Such parameters no longer need to be explicitly declared and passed through in various functions like `pull_request_build`.

The current passing of parameters for existing guidelines has not yet been changed.  I could do that work in separate PRs if the package commiters encourage it.

This change was tested in the `guideline_linecounts_meet_thresholds` branch of my fork, which introduces a new guideline.  The unit tests ran to completion.  The unit test for the new guideline succeeded.


MarkNahabedian:
Huh?  Is there something I need to do?

I think this PR is awaiting review.

DilumAluthge:
@ericphanson and/or @GunnarFarneback Can you review this?

PR #511, opened 2023-06-10
Alphanumeric name check: Don't use `\w`, because we don't want to include underscores

DilumAluthge:
This is a bug-fix. This guideline was always intended to exclude non-alphanumeric characters; we just weren't doing that correctly.


DilumAluthge:
H/t @GunnarFarneback for spotting this regex issue on Slack.

GunnarFarneback (approved):

DilumAluthge:
bors merge

PR #516, opened 2023-10-06
Require `[compat]` for stdlib dependencies

### LINKED ISSUE #515: Require compat statements for upgradeable stdlibs.
We currently have two upgradeable stdlibs, `Statistics` and `DelimitedFiles`.

We are looking at many more, but it opened the question, if and how we are enforcing that new package versions declare compatibility requirements with them.

I would almost prefer if we start requiring compat declarations for all standard libraries, so that the migration to upgradeable is smoother.

Right now there might be a window where we open something like https://github.com/JuliaRegistries/General/pull/92622 and a new package version is merged in between without declaring compat on LazyArtifacts.

vchuravy:
Fixes #515, by no longer treating stdlibs as special.

---

(Note: previously, this PR also added the requirement for JLL dependencies. That was removed from this PR.)


vchuravy:
This may be to aggressive, but I do think that we need compat entries at least for every stdlib.

DilumAluthge (approved):
LGTM.

Note that the JLL requirement will only apply to non-JLL packages. In other words:
1. Suppose that Foo.jl depends on Bar.jl and Hello_jll.jl. Foo.jl must have a compat entry for Bar.jl (this was already the case before this PR). *Foo.jl must have a compat entry for Hello_jll.jl (this is new, introduced in this PR).*
2. Suppose that Hello_jll.jl depends on World_jll.jl. Hello_jll.jl does NOT need a compat entry for World_jll.jl. This is one of the "special JLL exceptions" that we apply to JLL packages that come from Yggdrasil.

DilumAluthge:
cc: @staticfloat @giordano (to take a quick look at the JLL-related changes)

ericphanson:
This change makes sense to me, but when deployed, this should be accompanied by a discourse post announcing the change and the rationale

vchuravy:
@ericphanson how about the following annoucement

# PSA: Compat requirements in the General registry are changing.

Up until now standard libraries (and JLL's) have been excluded from the mandatory
requirement of being listed in `[compat]`. Since the version of a standard library was
tightly coupled to the version of Julia it was shipped with.

In Julia v1.11 (and to an lesser extend v1.10) we are moving towards **upgradeable** standard libaries,
this means that Julia will still come with the same selection of standard libraries, but instead of waiting
for a new Julia release, standard libraries will have their own release cycle. So it should be possible in 
the future to use Pkg v1.12 on Julia v1.11.

In effect standard libraries are becoming normal Julia packages, with the only difference that a version of
them will be shipped with a default Julia install.

Now this means that version numbers of standard libraries and Julia are no longer coupled and one
must manually declare `[compat]` requirements for standard libraries as well as Julia in your `Project.toml`. 
We will handle the registry side of things automatically, but developers will have to
manually adjust their `Project.toml` the next time they release a version to the registry.

So a Project.toml like:

```
name = "TestStdlibMissingCompat"
uuid = "4a0d4eb6-ee8c-4d0f-9d49-1daa6f6c32cf"
authors = ["Valentin Churavy <v.churavy@gmail.com>"]
version = "0.1.0"

[deps]
Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"

[compat]
julia = "1.9"
```

Would become:

```
```
name = "TestStdlibMissingCompat"
uuid = "4a0d4eb6-ee8c-4d0f-9d49-1daa6f6c32cf"
authors = ["Valentin Churavy <v.churavy@gmail.com>"]
version = "0.1.0"

[deps]
Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"

[compat]
julia = "1.9"
Statistics = "1.9"
```

**Comment by DilumAluthge** [comment]
In your example `Project.toml` file, can you also add an example JLL dependency, just for purposes of illustration?

**Review by giordano — COMMENTED** [review]
If this doesn't enforce compat bounds in jlls, as said above, then it looks good to me. For enforcing compat bounds also in jlls, which I'd like to, I'd need help with https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/314, I had troubles with installation of stdlib jlls in the build environment

**Comment by ericphanson** [comment]
That sounds great to me, but I agree with Dilum we should mention more JLLs too, not just stdlibs, since we're changing both here. (We mentioned they were exempt but not that they are no longer exempt). Also as a tiny grammar nit,

> Since the version of a standard library was tightly coupled to the version of Julia it was shipped with.

is kinda an incomplete sentence; maybe:

> Stdlibs were exempt because the version of a standard library was tightly coupled to the version of Julia it was shipped with.

**Comment by DilumAluthge** [comment]
@vchuravy and/or @staticfloat - can you help @giordano out with https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/314?

File: src/AutoMerge/guidelines.jl [hunk 1]
```diff
@@ -91,8 +91,8 @@ function meets_compat_for_all_deps(working_directory::AbstractString, pkg, versi
     for version_range in keys(deps)
         if version in Pkg.Types.VersionRange(version_range)
             for name in keys(deps[version_range])
-                if !is_jll_name(name) && !is_julia_stdlib(name)
-                    @debug("Found a new (non-stdlib non-JLL) dependency: $(name)")
+                if !is_jll_name(name)
+                    @debug("Found a new dependency: $(name)")
```
DilumAluthge, line 95: Probably want to restore the "non-JLL" portion of this log message.

vchuravy:
After a conversation with @DilumAluthge I would say let's back out the JLL change for now, that seems like a bigger change and the stdlib change needs to happen now.
We are already likely to have some inconsistencies in the Registry.

DilumAluthge:
https://discourse.julialang.org/t/psa-compat-requirements-in-the-general-registry-are-changing/104958?u=dilumaluthge

DilumAluthge:
Let's call this breaking. I've bumped the major version number to `9`.

DilumAluthge:
@vchuravy We need to fix several of the integration tests.

DilumAluthge:
bors try

DilumAluthge (approved):
LGTM

DilumAluthge:
bors merge

DilumAluthge:
This will be registered in https://github.com/JuliaRegistries/General/pull/94135.

DilumAluthge:
This will be deployed by the combination of https://github.com/JuliaRegistries/General/pull/94141 and https://github.com/JuliaRegistries/General/pull/94142

vchuravy:
I reran StdlibRegistryCompatUpdater for DelimitedFiles and Statistics https://github.com/JuliaRegistries/General/pull/94159

PR #524, opened 2023-10-28
When yanking a version, only `yanked = true` is correct; `yank = true` (and other variants) are incorrect, so we should detect them

### LINKED ISSUE #523: Check for misspelling of `yanked` as `yank`?
`yanked` is correct.

`yank` (or any other variant) is incorrect.

DilumAluthge:
Fixes #523


DilumAluthge:
bors merge

PR #525, opened 2023-10-28
Registry consistency tests: For every version, check that each (weak)compat entry has a corresponding (weak)deps entry

### LINKED ISSUE #522: Invariance check that could be added
https://github.com/JuliaRegistries/General/pull/93498 is buggy but passed CI. The missing check is that for every version, each compat entry needs to have a corresponding entry with that name in `Deps.toml`.

DilumAluthge:
Fixes #522


DilumAluthge:
bors try

DilumAluthge:
bors try

src/registry_testing.jl
```diff
@@ -238,6 +238,32 @@ function test(path=pwd(); registry_deps::Vector{<:AbstractString}=String[])
                 else
                     @debug "Compat.toml file does not exist" compatfile
                 end
+
+                # https://github.com/JuliaRegistries/RegistryCI.jl/issues/522
+                # For every version, each compat entry has a corresponding deps entry.
+                #
+                # Note: it is legal for a package to have Compat.toml but not have
+                # Deps.toml. Specifically, this is legal if (and only if) the package
+                # does not have any dependencies, and does have a compat entry for `julia`.
```
GunnarFarneback, line 247: Shouldn't WeakDeps.toml be part of this reasoning too?

DilumAluthge: Yeah, we should do the same thing for WeakDeps/WeakCompat.

DilumAluthge:
bors try

src/registry_testing.jl
DilumAluthge: Done. We now do this for both Deps/Compat and also for WeakDeps/WeakCompat.

For Deps/Compat, it is legal for Compat.toml to exist but Deps.toml to not exist - specifically, if a package has no dependencies, and it has a compat entry for `julia`, then it will have Compat.toml but not Deps.toml.

However, if I understand correctly, it is never legal for a package to have Compat.toml but not have Deps.toml.

DilumAluthge:
@GunnarFarneback or @ericphanson Would you be able to review this?

src/registry_testing.jl
DilumAluthge: I think that this has been resolved by https://github.com/JuliaRegistries/RegistryCI.jl/pull/525/commits/631ed08338e912a4e921e33a4bfbd03d942c9a1e

ericphanson:
#605 adds some registry unit tests with MetaTesting.jl; we should rebase this PR on top of that one and add additional unit tests to cover this new code

PR #536, opened 2023-12-09
AutoMerge: Respect the `Override AutoMerge: name similarity is okay` label

ericphanson:
...by turning off the distance check whenever the label is applied. The github action should be updated to activate when labels are added, for best effect, as shown in the example yaml.

### DIFF
--- a/example_github_workflow_files/automerge.yml
+++ b/example_github_workflow_files/automerge.yml
@@ -10,6 +10,11 @@ on:
 
 jobs:
   AutoMerge:
+    # Run if the we are not triggered by a label OR we are triggered by a label, and that
+    # label is one that affects the execution of the workflow
+    # Note: since the label contains a colon, we need to use a workaround like https://github.com/actions/runner/issues/1019#issuecomment-810482716
+    # for the syntax to parse correctly.
+    if: "${{  github.event.action != 'labeled' || (github.event.action == 'labeled' && github.event.label.name == 'Override AutoMerge: name similarity is okay') }}"
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/example_github_workflow_files/automerge.yml
+++ b/example_github_workflow_files/automerge.yml
@@ -4,6 +4,8 @@ on:
   schedule:
     - cron: '05,17,29,41,53 * * * *'
   pull_request:
+    # Here we run when labels are applied, so that the "Override AutoMerge: name similarity is okay label is respected.
+    types: [labeled, synchronize]
   workflow_dispatch:
 
 jobs:
--- a/example_github_workflow_files/automerge.yml
+++ b/example_github_workflow_files/automerge.yml
@@ -4,6 +4,7 @@ on:
   schedule:
     - cron: '05,17,29,41,53 * * * *'
   pull_request:
+    # opened = run when the PR is first opened
     # labeled = run when labels are applied, so that the "Override AutoMerge: name similarity is okay label is respected.
     # synchronize = run when a commit is pushed to the PR
     types: [opened, labeled, synchronize]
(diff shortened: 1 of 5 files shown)


ericphanson:
bump @DilumAluthge

DilumAluthge:
Should we just rename our label (on the General repo) and remove the colon character?

ericphanson:
I think it's fine, it's nice to not have 2 labels or have to delete the old ones, and it just needs an extra quote

ericphanson:
Bump- would be nice if folks with triage access could help here!

example_github_workflow_files/automerge.yml
```diff
@@ -4,10 +4,17 @@ on:
   schedule:
     - cron: '05,17,29,41,53 * * * *'
   pull_request:
+    # Here we run when labels are applied, so that the "Override AutoMerge: name similarity is okay label is respected.
```
DilumAluthge, line 7:
```suggestion
    # labeled = run when labels are applied, so that the "Override AutoMerge: name similarity is okay label is respected.
    # synchronize = run when a commit is pushed to the PR
```

DilumAluthge: @ericphanson Does `synchronize` also cover when the PR is first opened, or do we also need to add `opened` to the list?

example_github_workflow_files/automerge.yml
```diff
@@ -4,10 +4,17 @@ on:
...
+    types: [labeled, synchronize]
   workflow_dispatch:
 
 jobs:
   AutoMerge:
+    # Run if the we are not triggered by a label OR we are triggered by a label, and that
+    # label is one that affects the execution of the workflow
+    # Note: since the label contains a colon, we need to use a workaround like https://github.com/actions/runner/issues/1019#issuecomment-810482716
+    # for the syntax to parse correctly.
+    if: "${{  github.event.action != 'labeled' || (github.event.action == 'labeled' && github.event.label.name == 'Override AutoMerge: name similarity is okay') }}"
```
DilumAluthge, line 17:
```suggestion
    if: "${{ github.event.action != 'labeled' || (github.event.action == 'labeled' && github.event.label.name == 'Override AutoMerge: name similarity is okay') }}"
```

DilumAluthge (changes requested):
I'm not thrilled by the fact that we now have to run AutoMerge when labels are added or removed. But, off the top of my head, I can't think of an alternate approach.

And I do think that having this functionality will reduce the amount of time that human registry committers will need to spend. So, let's give this a try.

I've made a few comments, and once those are addressed, I think this is good to go.

ericphanson:
> I'm not thrilled by the fact that we now have to run AutoMerge when labels are added or removed. But, off the top of my head, I can't think of an alternate approach.

We only run when some specific labels are added or removed, not any labels!

example_github_workflow_files/automerge.yml
ericphanson: Hm, I think we may need to add `opened` as well

example_github_workflow_files/automerge.yml
```diff
@@ -4,10 +4,17 @@ on:
   schedule:
     - cron: '05,17,29,41,53 * * * *'
   pull_request:
+    # Here we run when labels are applied, so that the "Override AutoMerge: name similarity is okay label is respected.
+    types: [labeled, synchronize]
```
ericphanson, line 8:
```suggestion
    types: [opened, labeled, synchronize]
```

example_github_workflow_files/automerge.yml
```diff
@@ -4,10 +4,18 @@ on:
   schedule:
     - cron: '05,17,29,41,53 * * * *'
   pull_request:
+    # labeled = run when labels are applied, so that the "Override AutoMerge: name similarity is okay label is respected.
```
DilumAluthge, line 7:
```suggestion
    # opened = run when the PR is first opened
    # labeled = run when labels are applied, so that the "Override AutoMerge: name similarity is okay label is respected.
```

DilumAluthge (changes requested):

DilumAluthge (approved):
Let's give this a try.

PR #537, opened 2023-12-31
add support for commit hash information

simonbyrne:
Part of https://github.com/JuliaLang/Pkg.jl/issues/3718.

This does not depend on any other changes, but the keys should match those in https://github.com/JuliaRegistries/RegistryTools.jl/pull/97


src/AutoMerge/guidelines.jl
```diff
@@ -619,17 +619,22 @@ const guideline_src_names_OK = Guideline(;
 )
 
 function meets_code_can_be_downloaded(registry_head, pkg, version, pr; pkg_code_path)
-    uuid, package_repo, subdir, tree_hash_from_toml = parse_registry_pkg_info(
+    uuid, package_repo, subdir, tree_hash_from_toml, commit_hash_from_toml, tag_name_from_toml = parse_registry_pkg_info(
```
DilumAluthge, line 622: Seems like now is as good a time as any to make the following change:

```suggestion
    (; uuid, package_repo, subdir, tree_hash_from_toml, commit_hash_from_toml, tag_name_from_toml) = parse_registry_pkg_info(
```

simonbyrne: i agree, but they aren't the names....

DilumAluthge: Ah, that's annoying.

simonbyrne: Also, this was only added in Julia 1.7.

PR #539, opened 2024-01-27
AutoMerge: Allow skipping sequential versions with author-approval criteria

ericphanson:
Based on #536 to avoid conflicts. That one should be merged first.

Requires https://github.com/JuliaRegistries/General/pull/98815

The flow here is:

* automerge comments, notes they can override
* package author comments "[merge approved]"
* GHA adds a label
* new label triggers automerge again
* automerge sees the new label and skips the sequential version check

FAQ:
* What happens if some one else comments "merge approved"? The label doesn't get applied, so nothing happens
* What happens if someone adds the label manually? The label does take effect. I.e. triage-level maintainers can facilitate author approval if needed. We need to trust them to not abuse this (I doubt it will happen and we can revoke triage access if folks aren't behaving)
* Does a "merge approved" comment block automerge unless it contains `[noblock]`? No, we special case it so comments that trigger author-approval count as noblock.
* Does AutoMerge's comment mentioning "merge approved" trigger the labelling workflow? No, it shouldn't, it is a bot user. (If it did, the label would still not be applied, since the registrator is not the package author)
* What happens if multiple guidelines use the author-approval mechanism, does the author have to indicate what they are approving? No, `merge approved` will apply to all of them
* What happens if the author comments before even seeing the automerge comment? They would end up approving whatever it says, but only 1 guideline is overridable at this point.
* What happens if the PR is from gitlab? We won't be able to find the author's github username in the PR text from Registrator, so the labeling workflow won't be able to check the commenter is the author, so the label won't get added. Triage-level maintainers can manually add the label in this case, but there isn't an automatic solution.
* What happens if the PR is not from Registrator? The label could still get added by the workflow (if it hits the regex to have the author name, and the person with that username comments), but automerge won't run, so the label won't do anything.


DilumAluthge:
> * GHA adds a label
> * new label triggers automerge again

This might be a problem. If the label was added using the default `GITHUB_TOKEN` token, then IIRC the "label-added" event won't trigger any further GitHub Actions.

DilumAluthge:
> > * GHA adds a label
> > * new label triggers automerge again
> 
> This might be a problem. If the label was added using the default `GITHUB_TOKEN` token, then IIRC the "label-added" event won't trigger any further GitHub Actions.

Can we make the "GHA adds a label" step use the custom `@JuliaTagBot` PAT we have? I believe that this would bypass the above issue.

Of course, maybe we should first verify that the above actually would be an issue.

DilumAluthge:
Perhaps we should make the text specific to the guideline being approved? How about something like `/approve-merge-skip-version`?

Also, how about we add the `/` to the front, to emphasize that this is kind of a "command" that the package author is providing?

DilumAluthge:
Even if we only get this working for GitHub-hosted packages at first, this still covers the majority of packages, so I still think it'll help reduce the manual registry-committer burden.

ericphanson:
> This might be a problem. If the label was added using the default GITHUB_TOKEN token, then IIRC the "label-added" event won't trigger any further GitHub Actions.

Good point, I think that sounds right.

> Can we make the "GHA adds a label" step use the custom @JuliaTagBot PAT we have? I believe that this would bypass the above issue.
> Of course, maybe we should first verify that the above actually would be an issue.

Yeah, that makes sense. Since this can be done on the GHA by itself, IMO we can work on that once we merge down the stuff here and get RegistryCI updated on General. Then we can try it out and get the authentication working there.

> Perhaps we should make the text specific to the guideline being approved? How about something like /approve-merge-skip-version?
> Also, how about we add the / to the front, to emphasize that this is kind of a "command" that the package author is providing?

To me this seems overly complicated. Since RegistryCI prints all of the problems it found at once, the author can either approve merging or not knowing that information. I don't really feel like there's more than 1 decision the author actually has, since not approving 1 thing blocks merging (so ultimately they need to decide if they want to merge now or not).

I also think this will lead to more questions/issues where folks approve 1 thing but not another and are confused.

I think "merge approved" is straightforward and simpler.

> Even if we only get this working for GitHub-hosted packages at first, this still covers the majority of packages, so I still think it'll help reduce the manual registry-committer burden.

Agreed!

ericphanson:
Re- the token, the [docs here](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow) seem clear that indeed `GITHUB_TOKEN` is insufficient for this, and we should use a PAT. We could use an existing one or make a new one for the task.

DilumAluthge:
> We could use an existing one or make a new one for the task.

For now, probably easier to re-use the existing one (the `@JuliaTagBot` token that AutoMerge uses when merging a PR). But we might consider making a new one in the future.

src/AutoMerge/guidelines.jl
```diff
@@ -461,6 +461,10 @@ function _valid_change(old_version::VersionNumber, new_version::VersionNumber)
     end
 end
 
+const AUTHOR_APPROVAL_INSTRUCTIONS = string(
+    "**If this was not a mistake and you wish to merge this PR anyway,",
+    "write a comment that says \"merge approved\".**")
```
DilumAluthge, line 466:
```suggestion
    "write a comment that says `[merge approved]`.**")
```

DilumAluthge (changes requested):

src/AutoMerge/cron.jl
```diff
@@ -55,7 +55,11 @@ function all_specified_check_runs_passed(
     return all(values(check_passed))
 end
 
-pr_comment_is_blocking(c::GitHub.Comment) = !occursin("[noblock]", body(c))
+function pr_comment_is_blocking(c::GitHub.Comment)
+    # Note: `merge_approved` is not case sensitive, to match the semantics of `contains` on GHA
+    not_blocking = occursin("[noblock]", body(c)) || occursin("merge approved", lowercase(body(c)))
```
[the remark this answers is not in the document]
DilumAluthge: As discussed in https://github.com/JuliaRegistries/General/pull/98815, we're going to compromise, and we'll require `[merge approved]`, but not require `[noblock]`.

src/AutoMerge/util.jl
```diff
@@ -288,3 +288,19 @@ function get_all_non_jll_package_names(registry_dir::AbstractString)
     unique!(packages)
     return packages
 end
+
+const AUTHOR_APPROVED_LABEL = "package-author-approved"
```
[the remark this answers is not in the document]
ericphanson:
```suggestion
const AUTHOR_APPROVED_LABEL = "Override AutoMerge: package author approved"
```

test/automerge-unit.jl
```diff
@@ -129,6 +129,19 @@ end
         @test !AutoMerge.perform_distance_check([GitHub.Label(; name="Override AutoMerge: name similarity is okay")])
         @test !AutoMerge.perform_distance_check([GitHub.Label(; name="hi"), GitHub.Label(; name="Override AutoMerge: name similarity is okay")])
     end
+    @testset "has_author_approved_label" begin
+        @test !AutoMerge.has_author_approved_label(nothing)
+        @test !AutoMerge.has_author_approved_label([GitHub.Label(; name="hi")])
+        @test AutoMerge.has_author_approved_label([GitHub.Label(; name="package-author-approved")])
```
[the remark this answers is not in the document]
ericphanson:
```suggestion
        @test AutoMerge.has_author_approved_label([GitHub.Label(; name="Override AutoMerge: package author approved")])
```

src/AutoMerge/guidelines.jl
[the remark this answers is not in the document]
DilumAluthge: We will do https://github.com/JuliaRegistries/RegistryCI.jl/pull/539#discussion_r1468970719 instead.

test/automerge-unit.jl
```diff
@@ -129,6 +129,19 @@ end
...
+        @test AutoMerge.has_author_approved_label([GitHub.Label(; name="hi"), GitHub.Label(; name="package-author-approved")])
```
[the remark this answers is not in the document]
ericphanson:
```suggestion
        @test AutoMerge.has_author_approved_label([GitHub.Label(; name="hi"), GitHub.Label(; name="Override AutoMerge: package author approved")])
```

src/AutoMerge/guidelines.jl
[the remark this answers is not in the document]
ericphanson:
```suggestion
    "write a comment that contains the text: `[merge approved]`.**")
```

example_github_workflow_files/automerge.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     # label is one that affects the execution of the workflow
     # Note: since the label contains a colon, we need to use a workaround like https://github.com/actions/runner/issues/1019#issuecomment-810482716
     # for the syntax to parse correctly.
-    if: "${{  github.event.action != 'labeled' || (github.event.action == 'labeled' && github.event.label.name == 'Override AutoMerge: name similarity is okay') }}"
+    if: "${{  github.event.action != 'labeled' || (github.event.action == 'labeled' && (github.event.label.name == 'Override AutoMerge: name similarity is okay' || github.event.label.name == 'package-author-approved')) }}"
```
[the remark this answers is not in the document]
ericphanson:
```suggestion
    if: "${{  github.event.action != 'labeled' || (github.event.action == 'labeled' && (github.event.label.name == 'Override AutoMerge: name similarity is okay' || github.event.label.name == 'Override AutoMerge: package author approved')) }}"
```

test/automerge-unit.jl
```diff
@@ -129,6 +129,19 @@ end
...
+        @test !AutoMerge.has_author_approved_label([GitHub.Label(; name="hi")])
+        @test AutoMerge.has_author_approved_label([GitHub.Label(; name="Override AutoMerge: package author approved")])
+        @test AutoMerge.has_author_approved_label([GitHub.Label(; name="hi"), GitHub.Label(; name="Override AutoMerge: package author approved")])
+    end
+    @testset "pr_comment_is_blocking" begin
+        @test AutoMerge.pr_comment_is_blocking(GitHub.Comment(; body="hi"))
+        @test AutoMerge.pr_comment_is_blocking(GitHub.Comment(; body="block"))
+        @test !AutoMerge.pr_comment_is_blocking(GitHub.Comment(; body="[noblock]"))
+        @test !AutoMerge.pr_comment_is_blocking(GitHub.Comment(; body="[noblock]hi"))
+        @test !AutoMerge.pr_comment_is_blocking(GitHub.Comment(; body="merge approved abc"))
```
ericphanson, line 143:
```suggestion
        @test !AutoMerge.pr_comment_is_blocking(GitHub.Comment(; body="[merge approved] abc"))
```

docs/src/guidelines.md
```diff
@@ -23,6 +23,7 @@ function guidelines_to_markdown_output(guidelines_function::Function)
         this_is_jll_package = false,
         this_pr_can_use_special_jll_exceptions = false,
         use_distance_check = false,
+        author_approved = false
```
DilumAluthge, line 26:
```suggestion
        author_approved = false,
```

docs/src/guidelines.md
```diff
@@ -52,6 +53,7 @@ function guidelines_to_markdown_output(guidelines_function::Function)
...
```
DilumAluthge, line 56:
```suggestion
        author_approved = false,
```

src/AutoMerge/cron.jl
DilumAluthge, line 60:
```suggestion
    not_blocking = occursin("[noblock]", body(c)) || occursin("[merge approved]", lowercase(body(c)))
```

DilumAluthge (changes requested):

DilumAluthge:
@ericphanson This needs a rebase, I think.

ericphanson:
(updated)

example_github_workflow_files/automerge.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     # label is one that affects the execution of the workflow
     # Note: since the label contains a colon, we need to use a workaround like https://github.com/actions/runner/issues/1019#issuecomment-810482716
     # for the syntax to parse correctly.
-    if: "${{ github.event.action != 'labeled' || (github.event.action == 'labeled' && github.event.label.name == 'Override AutoMerge: name similarity is okay') }}"
+    if: "${{  github.event.action != 'labeled' || (github.event.action == 'labeled' && (github.event.label.name == 'Override AutoMerge: name similarity is okay' || github.event.label.name == 'Override AutoMerge: package author approved')) }}"
```
DilumAluthge, line 19:
```suggestion
    if: "${{ github.event.action != 'labeled' || (github.event.action == 'labeled' && (github.event.label.name == 'Override AutoMerge: name similarity is okay' || github.event.label.name == 'Override AutoMerge: package author approved')) }}"
```

DilumAluthge (changes requested):

docs/src/guidelines.md
```diff
@@ -23,6 +23,7 @@ function guidelines_to_markdown_output(guidelines_function::Function)
         this_is_jll_package = false,
         this_pr_can_use_special_jll_exceptions = false,
         use_distance_check = false,
+        author_approved = false,
```
DilumAluthge, line 26:
```suggestion
        package_author_approved = false,
```

docs/src/guidelines.md
```diff
@@ -52,6 +53,7 @@ function guidelines_to_markdown_output(guidelines_function::Function)
...
```
DilumAluthge, line 56:
```suggestion
        package_author_approved = false,
```

src/AutoMerge/cron.jl
```diff
@@ -55,7 +55,11 @@ function all_specified_check_runs_passed(
     return all(values(check_passed))
 end
 
-pr_comment_is_blocking(c::GitHub.Comment) = !occursin("[noblock]", body(c))
+function pr_comment_is_blocking(c::GitHub.Comment)
+    # Note: `merge_approved` is not case sensitive, to match the semantics of `contains` on GHA
```
DilumAluthge, line 59:
```suggestion
    # Note: `[merge approved]` is not case sensitive, to match the semantics of `contains` on GitHub Actions
```

src/AutoMerge/guidelines.jl
```diff
@@ -976,7 +981,8 @@ function get_automerge_guidelines(
     check_license::Bool,
     this_is_jll_package::Bool,
     this_pr_can_use_special_jll_exceptions::Bool,
-    use_distance_check::Bool
+    use_distance_check::Bool,
+    author_approved::Bool # currently unused for new packages
```
DilumAluthge, line 985:
```suggestion
    package_author_approved::Bool # currently unused for new packages
```

src/AutoMerge/guidelines.jl
```diff
@@ -1018,12 +1024,13 @@ function get_automerge_guidelines(
     check_license::Bool,
     this_is_jll_package::Bool,
     this_pr_can_use_special_jll_exceptions::Bool,
-    use_distance_check::Bool # unused for new versions
+    use_distance_check::Bool, # unused for new versions
+    author_approved::Bool
```
DilumAluthge, line 1028:
```suggestion
    package_author_approved::Bool,
```

src/AutoMerge/guidelines.jl
```diff
@@ -1018,12 +1024,13 @@ function get_automerge_guidelines(
...
 )
     guidelines = [
         (guideline_registry_consistency_tests_pass, true),
         (guideline_pr_only_changes_allowed_files, true),
-        (guideline_sequential_version_number, !this_pr_can_use_special_jll_exceptions),
+        (guideline_sequential_version_number, !this_pr_can_use_special_jll_exceptions && !author_approved),
```
DilumAluthge, line 1033:
```suggestion
        (guideline_sequential_version_number, !this_pr_can_use_special_jll_exceptions && !package_author_approved),
```

src/AutoMerge/pull_requests.jl
```diff
@@ -195,7 +195,8 @@ function pull_request_build(data::GitHubAutoMergeData; check_license)::Nothing
         check_license=check_license,
         this_is_jll_package=this_is_jll_package,
         this_pr_can_use_special_jll_exceptions=this_pr_can_use_special_jll_exceptions,
-        use_distance_check=perform_distance_check(data.pr.labels)
+        use_distance_check=perform_distance_check(data.pr.labels),
+        author_approved=has_author_approved_label(data.pr.labels)
```
DilumAluthge, line 199:
```suggestion
        package_author_approved=has_package_author_approved_label(data.pr.labels)
```

src/AutoMerge/util.jl
```diff
@@ -288,3 +288,19 @@ function get_all_non_jll_package_names(registry_dir::AbstractString)
     unique!(packages)
     return packages
 end
+
+const AUTHOR_APPROVED_LABEL = "Override AutoMerge: package author approved"
```
DilumAluthge, line 292:
```suggestion
const PACKAGE_AUTHOR_APPROVED_LABEL = "Override AutoMerge: package author approved"
```

src/AutoMerge/util.jl
```diff
@@ -288,3 +288,19 @@ function get_all_non_jll_package_names(registry_dir::AbstractString)
...
+
+function has_author_approved_label(labels)
```
DilumAluthge, line 294:
```suggestion
function has_package_author_approved_label(labels)
```

src/AutoMerge/util.jl
```diff
@@ -288,3 +288,19 @@ function get_all_non_jll_package_names(registry_dir::AbstractString)
...
+    # No labels? Not approved
+    isnothing(labels) && return false
+    for label in labels
+        if label.name === AUTHOR_APPROVED_LABEL
```
DilumAluthge, line 298:
```suggestion
        if label.name === PACKAGE_AUTHOR_APPROVED_LABEL
```

src/AutoMerge/util.jl
```diff
@@ -288,3 +288,19 @@ function get_all_non_jll_package_names(registry_dir::AbstractString)
...
+            # found the approval
+            @debug "Found `$(AUTHOR_APPROVED_LABEL)` label"
```
DilumAluthge, line 300:
```suggestion
            @debug "Found `$(PACKAGE_AUTHOR_APPROVED_LABEL)` label"
```

src/AutoMerge/guidelines.jl
```diff
@@ -461,6 +461,10 @@ function _valid_change(old_version::VersionNumber, new_version::VersionNumber)
     end
 end
 
+const AUTHOR_APPROVAL_INSTRUCTIONS = string(
```
DilumAluthge, line 464:
```suggestion
const PACKAGE_AUTHOR_APPROVAL_INSTRUCTIONS = string(
```

src/AutoMerge/guidelines.jl
```diff
@@ -471,6 +475,7 @@ const guideline_sequential_version_number = Guideline(;
         "valid new versions are `1.0.1`, `1.1.1`, `1.2.0` and `2.0.0`. ",
         "Invalid new versions include `1.0.2` (skips `1.0.1`), ",
         "`1.3.0` (skips `1.2.0`), `3.0.0` (skips `2.0.0`) etc.",
+        AUTHOR_APPROVAL_INSTRUCTIONS,
```
DilumAluthge, line 478:
```suggestion
        PACKAGE_AUTHOR_APPROVAL_INSTRUCTIONS,
```

test/automerge-unit.jl
```diff
@@ -129,6 +129,19 @@ end
         @test !AutoMerge.perform_distance_check([GitHub.Label(; name="Override AutoMerge: name similarity is okay")])
         @test !AutoMerge.perform_distance_check([GitHub.Label(; name="hi"), GitHub.Label(; name="Override AutoMerge: name similarity is okay")])
     end
+    @testset "has_author_approved_label" begin
+        @test !AutoMerge.has_author_approved_label(nothing)
```
DilumAluthge, line 133:
```suggestion
        @test !AutoMerge.has_package_author_approved_label(nothing)
```

test/automerge-unit.jl
```diff
@@ -129,6 +129,19 @@ end
...
+        @test !AutoMerge.has_author_approved_label([GitHub.Label(; name="hi")])
```
DilumAluthge, line 134:
```suggestion
        @test !AutoMerge.has_package_author_approved_label([GitHub.Label(; name="hi")])
```

test/automerge-unit.jl
```diff
@@ -129,6 +129,19 @@ end
...
+        @test AutoMerge.has_author_approved_label([GitHub.Label(; name="Override AutoMerge: package author approved")])
```
DilumAluthge, line 135:
```suggestion
        @test AutoMerge.has_package_author_approved_label([GitHub.Label(; name="Override AutoMerge: package author approved")])
```

test/automerge-unit.jl
```diff
@@ -129,6 +129,19 @@ end
...
+        @test AutoMerge.has_author_approved_label([GitHub.Label(; name="hi"), GitHub.Label(; name="Override AutoMerge: package author approved")])
```
DilumAluthge, line 136:
```suggestion
        @test AutoMerge.has_package_author_approved_label([GitHub.Label(; name="hi"), GitHub.Label(; name="Override AutoMerge: package author approved")])
```

PR #541, opened 2024-01-29
add "List of all GitHub PR labels that can influence AutoMerge"

ericphanson:

PR #542, opened 2024-01-29
Bump version for release

ericphanson:
Also updates tagbot, so we have the best chances of actually getting a tag.

When we register this, we can include these release notes:

```
* Turn off stdlib compat requirement ([#540](https://github.com/JuliaRegistries/RegistryCI.jl/pull/540))
* Adds label-reading support for author-approval workflow ([#539](https://github.com/JuliaRegistries/RegistryCI.jl/pull/539)) and overriding the name similarity check ([#536](https://github.com/JuliaRegistries/RegistryCI.jl/pull/536))
```

We should only tag after landing #539 and #541

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -4,6 +4,22 @@ on:
     types:
       - created
   workflow_dispatch:
+    inputs:
+      lookback:
+        default: 3
+permissions:
+  actions: read
+  checks: read
+  contents: write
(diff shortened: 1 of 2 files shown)


ericphanson:
It's in there, in the second bullet

DilumAluthge:
Ah, I did not scroll to the right.

DilumAluthge (approved):

PR #547, opened 2024-01-30
Add reference test for AutoMerge comment (and all its permutations)

ericphanson:
This is a test-only change that adds ReferenceTests as a test-dep and uses it to test the AutoMerge comment. What this allows us to do is surface changes to the generated comment in PR diffs. This is a first step at #150.

When the comment changes and you run tests interactively, ReferenceTests.jl will provide you an easy way to update the references (and fail the tests). If they are not updated, CI will fail, so we can be sure they are up-to-date.


ericphanson:
nightly CI failure is unrelated; not sure if it will let me merge though

DilumAluthge:
> nightly CI failure is unrelated; not sure if it will let me merge though

It should let you merge.

PR #551, opened 2024-01-30
split out `guideline_name_match_check`

ericphanson:
This part should not be overridable by labels. (Note this isn't a huge issue, since the consistency tests seem to catch this already, but it is helpful to have the logic clear here and to give a good message even if the label is applied removing the similarity check).

### DIFF
--- a/src/AutoMerge/guidelines.jl
+++ b/src/AutoMerge/guidelines.jl
@@ -264,6 +264,7 @@ function sqrt_normalized_vd(name1, name2)
     return VisualStringDistances.visual_distance(name1, name2; normalize=x -> 5 + sqrt(x))
 end
 
+# This check cannot be overridden, since it's important for registry integrity
 const guideline_name_match_check = Guideline(;
     info = "Name does not match the name of any existing package names (up-to-case)",
     docs = "Packages must not match the name of existing package up-to-case, since on case-insensitive filesystems, this will break the registry.",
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Can you add a comment somewhere saying explicitly that this check is not label-overridable? (I quickly skimmed the diff but didn't see such a comment.)

src/AutoMerge/guidelines.jl
```diff
@@ -302,18 +331,9 @@ function meets_distance_check(
...
-            push!(
-                problem_messages,
-                (
-                    "Package name matches existing package name $(other_pkg) up to case.",
-                    (0, 0, 0),
-                ),
-            )
+        if lowercase(pkg_name) == lowercase(other_pkg)
+            # We handle this case in `meets_name_match_check`
+            continue
```
DilumAluthge, line 336: It seems like it would be fine to just return `false` here, right? Instead of continuing?

```suggestion
            # We can short-circuit in this case.
            return (false, "Package name matches existing package name $(other_pkg) up to case.")
```

DilumAluthge (approved):

src/AutoMerge/guidelines.jl
ericphanson: I think it's still nice to show the list of other problems, because it helps tell them what other similar names also exist that they shouldn't use (since they will have to change their name already and might want to choose something similar)

DilumAluthge: Oh I see. If we short-circuit now, then they don't get the full list of other name similarities.

PR #555, opened 2024-02-05
Don't automerge packages whose names end with jl

LilithHafner:
Inspired by https://github.com/JuliaRegistries/General/pull/100209. Whether or not folks decide to allow the name PsychoJL.jl, all such names should require manual approval.

### DIFF
--- a/src/AutoMerge/guidelines.jl
+++ b/src/AutoMerge/guidelines.jl
@@ -244,7 +244,7 @@ function meets_name_ascii(pkg)
 end
 
 const guideline_julia_name_check = Guideline(;
-    info="Name does not include \"julia\" or start with \"Ju\".",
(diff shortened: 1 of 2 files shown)


ericphanson:
BTW IMO we can test it with [unit tests for `meets_julia_name_check`](https://github.com/JuliaRegistries/RegistryCI.jl/blob/ee1d7cdb165202f4f3929a122c3188fbdd7afc16/test/automerge-unit.jl#L97) rather than full integration tests (which are more complicated)

ericphanson (approved):
lgtm!

ericphanson:
merge queue will run integration tests, and if they pass it will merge. We could bump for a release now as well to get it deployed, but I was hoping to get some more PRs in before the next release

PR #565, opened 2024-06-15
remove section numbers from AutoMerge comment

ericphanson:
* remove section numbers: distracting, not useful
* tweak titles: improve legibility a little

Sorry for all the churn here, but it helps to see it in action for a bit to realize what else could be improved.

To me this looks a little nicer, but I'm not 100% sure if it's actually better


ericphanson:
I don't think this is actually better

PR #566, opened 2024-06-25
try adding an "extended explanation" to compat bounds

ericphanson:
Adding an example Project.toml was suggested on Slack by @leonbello to help newcomers see exactly what we're talking about in the message. Here, I've tucked it under an "extended explanation" but we could also just have much longer explanations in the main text.

See [here](https://github.com/JuliaRegistries/RegistryCI.jl/blob/eph/extended/test/reference_comments/comment_pass_false_type_new_package_suggest_onepointzero_false_version_0.1.0_point_to_slack_false.md) for an example of how it looks


ericphanson:
I think this is an improvement on the status quo, so I'll merge and tag but I'm happy to continue adjusting this text or style in followup PRs if there's post-merge feedback.

DilumAluthge:
LGTM.

DilumAluthge:
In addition to linking to the Pkg docs, can we also link to https://juliaregistries.github.io/RegistryCI.jl/stable/guidelines/#Upper-bounded-%5Bcompat%5D-entries?

PR #569, opened 2024-07-22
Open tagbot trigger issues for testitem repos

davidanthoff:
Repos that use the `julia-vscode/testitem-workflow/.github/workflows/juliaci.yml` are also TagBot enabled.

### DIFF
--- a/src/TagBot/TagBot.jl
+++ b/src/TagBot/TagBot.jl
@@ -71,7 +71,7 @@ function tagbot_file(repo; issue_comments=false)
         f.typ == "file" || continue
         file = GH.file(repo, f.path; auth=AUTH[])
         contents = String(base64decode(file.content))
-        if occursin("JuliaRegistries/TagBot", contents)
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Hmmm. So, IIUC, this code is looking for the text `JuliaRegistries/TagBot` in the _contents_ of the file (not the filename).

So I'm wondering why it's not already working, given that the https://github.com/julia-vscode/testitem-workflow/blob/main/.github/workflows/juliaci.yml file already includes the text `JuliaRegistries/TagBot` in the file contents.

DilumAluthge (approved):

ericphanson:
btw to be deployed, we need to first bump the version and merge & tag that

davidanthoff:
Should I be doing something to trigger that? Probably makes more sense for someone who is more in charge of this repo to do that?

ericphanson:
if it wasn't already merged I'd have asked you to include the version bump as part of the PR. Changes are infrequent currently so IMO it's better to just release whenever there's a change, otherwise it will be a long time until the next release.

PR #571, opened 2024-07-28
Allow LazyJLLWrappers for JLLs

staticfloat:
BinaryBuilder2 generates JLLs that use `LazyJLLWrappers` instead of `JLLWrappers`, we need to allow that here for them to get registered.

### DIFF
--- a/src/AutoMerge/jll.jl
+++ b/src/AutoMerge/jll.jl
@@ -31,17 +31,17 @@ const guideline_allowed_jll_nonrecursive_dependencies = Guideline(;
 function meets_allowed_jll_nonrecursive_dependencies(
     working_directory::AbstractString, pkg, version
 )
-    # If you are a JLL package, you are only allowed to have five kinds of dependencies:
+    # If you are a JLL package, you are only allowed to have the following dependencies:
     # 1. Pkg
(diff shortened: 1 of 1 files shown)


giordano (approved):
Looks good, as long as tests pass

DilumAluthge (approved):

staticfloat:
CI failed, but the failures don't look like my fault?

staticfloat:
Confirmed that the failures are not my fault, the same failures occur on a meaningless PR: https://github.com/JuliaRegistries/RegistryCI.jl/pull/572

DilumAluthge:
The problem is this line: https://github.com/JuliaBinaryWrappers/MozillaCACerts_jll.jl/blob/713bd3679eeb3671788eb5f3c5e5ca064a63cc99/Project.toml#L15

That line is incorrect. It should be `Libdl = "< 0.0.1, 1"` instead. Is this a bug in Ygg/BinaryBuilder?

giordano:
> Is this a bug in Ygg/BinaryBuilder?

Someone opened https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1324 but it's still marked as draft

PR #573, opened 2024-08-07
add `AutoMerge: last run blocked by comment` and `Override AutoMerge: ignore blocking comments` labels

ericphanson:
This is the first RegistryCI feature that involves adding/removing labels programatically, as opposed to previous work that only reads them. This only affects the cron job. When it runs, it

1. Checks if the label "AutoMerge: last run blocked by comment" exists. If not it creates it. This will require new permissions, so this change is likely breaking
2. For each PR, if the PR has the label "Override AutoMerge: ignore blocking comments" then the PR does not count as blocked
3. if a PR is blocked, the label "Override AutoMerge: ignore blocking comments" is added (again, this needs new permissions)
4. if a PR is not blocked, the label "Override AutoMerge: ignore blocking comments" is removed (again, permissions)

Specifically, the permissions needed are: the github token by the cron job needs at least one of:

* "Issues" repository permissions (write)
* "Pull requests" repository permissions (write)


DilumAluthge:
Unit tests on Julia 1.3 should be fixed now after https://github.com/JuliaRegistries/General/pull/112623. I'll retry that job.

ericphanson:
Is there a way to try the integration tests without merging?

DilumAluthge:
We could add a `workflow_dispatch` trigger to the integration tests, which would allow us to trigger then manually.

DilumAluthge:
All of this work is done in the cron job, right? This doesn't change anything in the PR job?

I haven't done a line-by-line review of the code, but at a high level, I understand the motivation here:
1. Increase the visibility (to the PR author) when a PR is blocked by blocking comments.
2. Give registry triagers the ability to override blocking comments.

ericphanson:
Yep, exactly. The “blocked” label is intended to be managed entirely by the cron job to communicate status, and is set and removed by it. If a human adds or removes it, it will just be corrected next cron run to surface the actual status. 

The override label is never set or removed by the cron job, only read. It allows ignoring blocking comments; it’s easier than editing “noblock” into every comment, and it allows triage members to overrule community blocks if necessary.

I hope these have the following effects:
* package authors can understand more easily when their PR is blocked (maybe even by their own comments)
* One can tell without even clicking into a PR which ones are blocked, or filter PRs by blocked/unblocked ones for an easier overview
* Triage members have a new tool for moderation/maintence that makes things slightly easier.

DilumAluthge:
Those details sound good to me.

DilumAluthge (approved):
I have not reviewed the code diff line-by-line, but the high-level plan sounds good to me.

ericphanson:
ok, think this is ready to go. I believe it is non-breaking, since while it needs write permissions for the PR, it is part of the cron job which needs merging permissions anyway. It does not add or remove labels in `read_only` mode.

DilumAluthge (approved):

test/automerge-integration.jl
```diff
@@ -283,6 +315,20 @@ hello_world_commit2 = "57b0aec49622faa962c6752d4bc39a62b91fe37c"
...
+                            labels = GitHub.labels(repo, pr)
+                            @test AutoMerge.has_label(labels, AutoMerge.BLOCKED_LABEL)
+                            # Delete the comment & rerun
+                            GitHub.delete_comment(repo, blocking_comment; auth=auth)
+                            sleep(1)
+                            merge()
+                            sleep(1)
+                            # Check we no longer have the blocked label
+                            labels = GitHub.labels(repo, pr)
+                            @test !AutoMerge.has_label(labels, AutoMerge.BLOCKED_LABEL)
```
ericphanson, line 330: this test is failing, but I'm not sure why. The label does seem to be deleted (https://github.com/bcbi-test/automerge-integration-test-repo/pull/11079)

CI:https://github.com/JuliaRegistries/RegistryCI.jl/actions/runs/11469400064/job/31916569056#step:8:1307

src/AutoMerge/cron.jl
```diff
@@ -384,7 +394,13 @@ function cron_or_api_build(
         return nothing
     end
 
-    if !pr_has_no_blocking_comments(api, registry, pr; auth=auth)
+    blocked = pr_has_blocking_comments(api, registry, pr; auth=auth) && !has_label(pr.labels, OVERRIDE_BLOCKS_LABEL)
+    if blocked
+        if !read_only
+            # add `BLOCKED_LABEL` to communicate to users
+            # that the PR is blocked from automerging
+            GitHub.add_labels(api, registry.full_name, pr_number, [BLOCKED_LABEL]; auth=auth)
```
giordano, line 402: Do I understand correctly this unconditionally adds the label even if it's there already?  What I'm seeing is that [PRs with this label](https://github.com/JuliaRegistries/General/pulls?page=2&q=sort%3Aupdated-desc+is%3Apr+is%3Aopen+label%3A%22AutoMerge%3A+last+run+blocked+by+comment%22) are "updated" every time automerge runs, even if the label is already there, with the result that these PRs will never be automatically closed as stale, as they're continuously updated every 10 minutes or so.  Also, this makes looking at the list of [recently updated PRs](https://github.com/JuliaRegistries/General/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen) hard, because these PRs are constantly bumped (and there are dozens of them, especially since they can't be automatically closed) and so there are dozens of them, covering other PRs that had been last updated mere minutes earlier than last automerge run.

PR #577, opened 2024-10-21
add time indication for "next round"

ericphanson:
> I think it would be useful to mention here what is the time interval between rounds ... to give people some idea of how much time they have to wait.

`@c05` on Slack


DilumAluthge:
Seems fine, as long as people understand that this is just a rough estimate (based on how often our stopwatch system triggers new cron jobs), and not an promise or a guarantee.

DilumAluthge (approved):

DilumAluthge (approved):

PR #578, opened 2024-10-25
Move blocked label updating earlier

ericphanson:
Currently, we don't update the labels regarding blocked/not-blocked unless the auto merge commit status is OK, the waiting period has elapsed, etc. This means many PRs that are blocked aren't labeled. It also means that the blocked label won't be properly removed in the following situation:

- the PR has automerge status passing, but a comment is blocking
- blocked label is added
- the pr is updated and automerge starts failing
- blocking comment is removed
- status quo: block label is *not* removed bc we don't get to the code that updates the label
- PR: block label is removed, since we update the block label first

This PR fixes it by moving the block check earlier, once we are confident it is a registration PR, so the label is updated reliably.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>", "contributors"]
-version = "10.8.0"
+version = "10.8.1"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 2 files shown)


DilumAluthge (approved):
I haven't reviewed the PR diff line-by-line, but based on Eric's description in the PR description, this makes sense to me.

ericphanson:
urg, I think it won't auto-merge since it's waiting for integration tests

DilumAluthge:
Hmmm I thought we fixed this already.

ericphanson:
ok, I was able to fix it by cancelling the in-progress integration tests from the actions tab, somehow that did the trick

DilumAluthge:
I just force-pushed, and now we are good to go - `Integration` is marked as `skipped` (not pending).

ericphanson:
ah or maybe it was something you did @DilumAluthge 😄

DilumAluthge:
Well either way, we have the correct result 😂

PR #582, opened 2024-12-08
[AutoMerge] Do not add "blocked" label if already there

giordano:
Untested, but I presume this should fix https://github.com/JuliaRegistries/RegistryCI.jl/pull/573#discussion_r1875021374.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RegistryCI"
 uuid = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>", "contributors"]
-version = "10.9.0"
+version = "10.9.1"
 
 [deps]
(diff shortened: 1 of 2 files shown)


ericphanson (approved):
good catch! I had assumed adding the label was idempotent and it was slightly better to unconditionally add it to avoid (somewhat unlikely) race conditions, such as due to `pr.labels` being stale. Did not realize it would bump the PR. Looks like a good fix to me. 

I wonder if there's anywhere else I'm not checking due to the same kind of assumptions?

giordano:
This seems to be [the only use of `add_labels`](https://github.com/search?q=repo%3AJuliaRegistries%2FRegistryCI.jl%20add_labels&type=code)

DilumAluthge:
Thank you both!

giordano:
For the record, [PRs with the "blocked" label](https://github.com/JuliaRegistries/General/pulls?page=2&q=sort%3Aupdated-desc+is%3Apr+is%3Aopen+label%3A%22AutoMerge%3A+last+run+blocked+by+comment%22) haven't been updated en masse for over an hour, more specifically since https://github.com/JuliaRegistries/General/pull/120968 was merged, so it'd appear this PR did its job.

PR #583, opened 2024-12-08
[CI] Some updates to GitHub Actions setup

giordano:
* Install dependabot
* Update various external workflows
* Use `julia-actions/cache` instead of `actions/cache`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/ci_integration.yml
+++ b/.github/workflows/ci_integration.yml
@@ -33,21 +33,12 @@ jobs:
     name: Integration
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: julia-actions/setup-julia@v1
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
         with:
           version: '1'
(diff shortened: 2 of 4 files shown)


DilumAluthge:
> * Use `julia-actions/setup-julia` instead of `actions/cache`

Presumably this should say `julia-actions/cache` instead of `julia-actions/setup-julia`?

DilumAluthge (approved):

ericphanson:
This one was waiting for integration tests to enter merge queue but they only run once in the merge queue. I think maybe we have to wait for #582 to merge and then try again here.

giordano:
> @github-merge-queue github-merge-queue bot removed this pull request from the merge queue due to no response for status checks

Well, the job is still running.....

DilumAluthge:
> > @github-merge-queue github-merge-queue bot removed this pull request from the merge queue due to no response for status checks
> 
> Well, the job is still running.....

This usually (in my experience) means that the GitHub Merge Queue settings (which are inside the branch protection settings for the `master` branch) needed to be edited; specifically, the Merge Queue timeout needs to be increased. This is the setting that is labeled as follows:

> Status check timeout
> Time (in minutes) a required status check must report a conclusion within to not be considered failed.
> Consider check failed after: 

Currently, that setting is at 120 minutes. I think 120 is sufficient for most cases, so I'll leave it unmodified for now. But if we keep seeing this error in the future, then we can increase it from 120 to something higher.

ericphanson:
I think 120 is insufficient, since our integration tests already take over an hour, and they have concurrency limit 1. What happened in this case was the integration tests were running for #582, and shortly after they started we added this PR to merge queue. After an hour, #582 was merged, and the integration tests started running here. Then we hit the 120 minute limit, then after that this PR's integration tests finished.

Since our integration tests take so long and can't run in parallel, I think 120 is too low, and it should be 240+

PR #586, opened 2024-12-16
add requirement and detection that breaking changes are explained in release notes

IanButterworth:
Helps get explanations of breaking changes into General PRs and ultimately TagBot-created releases.


DilumAluthge (commented):
Can we add a kwarg to the public interface (`AutoMerge.run()`) that allows users to opt-out of this guideline. I can imagine a situation where people are using AutoMerge internally for their private registry, and maybe they have a different internal system for NEWS, release notes, etc., and thus they don't want their internal AutoMerge to enforce a particular style.

ericphanson:
This may have coverage already, I've [kicked off](https://github.com/JuliaRegistries/RegistryCI.jl/actions/runs/12355612114) the integration tests (via workflow dispatch), we can see if they fail. If so we will need to figure out how to adjust them to get the release notes in there somehow.

The way integration tests work here is they run when the PR is in the merge queue, or by workflow dispatch by a maintainer.

IanButterworth:
Currently this only fails if the `BREAKING` label is applied. Is that sufficient?

IanButterworth:
@ericphanson sorry I think it needs re-invoking as there was a bug before.

ericphanson:
> Currently this only fails if the BREAKING label is applied. Is that sufficient?

I think this should be OK as registrator adds the label, so it _should_ be before RegistryCI queries github for the PR. I guess if Registrator is quite slow and RegistryCI is fast though, there is nothing guaranteeing that ordering. We might be safer if we query for the labels inside the guideline itself instead of relying on `pr.labels` which may be stale. However I think it's OK either way.

IanButterworth:
Makes sense. Updated.

src/AutoMerge/guidelines.jl
```diff
@@ -322,6 +322,35 @@ function meets_name_match_check(
...
 
+# This check checks for an explanation of why a breaking change is breaking
+const guideline_breaking_explanation = Guideline(;
+    info = "Release notes have not been provided that explain why this is a breaking change.",
+    docs = "If this is a breaking change, release notes must be given that explain why this is a breaking change (i.e. mention \"breaking\"). To update the release notes, please see the \"Providing and updating release notes\" subsection under \"Additional information\" below.",
+    check=data -> meets_breaking_explanation_check(data))
+
+function meets_breaking_explanation_check(data)
+    # Look up PR here in case the labels are slow to be applied by the Registrator bot
+    # which decides wheter to add the BREAKING label
```
ericphanson, line 333:
```suggestion
    # which decides whether to add the BREAKING label
```

ericphanson:
This looks good to me, the only thing left is @DilumAluthge's request [here](https://github.com/JuliaRegistries/RegistryCI.jl/pull/586#pullrequestreview-2506479380) and getting CI working

ericphanson (dismissed):
LGTM!
Can you bump the patch version? Then feel free to add to the merge queue and register.

On General's side, once its registered, to make use of it we will need to run the update-manifests workflow (https://github.com/JuliaRegistries/General/blob/master/.github/workflows/update_manifests.yml) then add the kwarg.

src/AutoMerge/guidelines.jl
```diff
@@ -322,6 +322,38 @@ function meets_name_match_check(
...
+    pr = GitHub.pull_request(data.api, data.registry, data.pr.number; auth=data.auth)
+    return meets_breaking_explanation_check(pr.labels, pr.body)
+end
+function meets_breaking_explanation_check(labels::Vector, body::AbstractString)
+    if any(==("BREAKING"), labels)
+        release_notes = get_release_notes(body)
+        if release_notes === nothing
+            return false, "This is a breaking change, but no release notes have been provided."
+        elseif !occursin(r"breaking", lowercase(release_notes))
+            return false, "This is a breaking change, but the release notes do not mention it."
```
ericphanson, line 343: BTW this might be one that could benefit from an extended explanation of how to do it, like https://github.com/JuliaRegistries/RegistryCI.jl/blob/f74fef6646903b00992b10a7b9e98f123c4762f9/src/AutoMerge/guidelines.jl#L93-L117

src/AutoMerge/guidelines.jl
```diff
@@ -322,6 +322,63 @@ function meets_name_match_check(
...
+function meets_breaking_explanation_check(data::GitHubAutoMergeData)
+    # Look up PR here in case the labels are slow to be applied by the Registrator bot
+    # which decides whether to add the BREAKING label
+    pr = GitHub.pull_request(data.api, data.registry, data.pr.number; auth=data.auth)
+    return meets_breaking_explanation_check(pr.labels, pr.body)
+end
+function meets_breaking_explanation_check(labels::Vector, body::AbstractString)
+    if any(==("BREAKING"), labels)
+        release_notes = get_release_notes(body)
+        example_detail = """
```
ericphanson, line 340: I'm a little worried about the indentation here, I remember that being a pain w/ the compat one, where detail block wouldn't render within the list itself

ericphanson: that one is https://github.com/JuliaRegistries/RegistryCI.jl/blob/master/test/reference_comments/comment_pass_false_type_new_package_suggest_onepointzero_false_version_1.0.0_point_to_slack_false.md#2-automerge-guidelines-which-are-not-met-

populated in the tests via https://github.com/JuliaRegistries/RegistryCI.jl/blob/11d89a142313fb8516b8351c57b3035956a8240a/test/automerge-unit.jl#L72

IanButterworth: I indented them. If that's not what you're saying, and it's clear what to do please feel free to push

ericphanson: sure, I pushed some commits, now it's in the reference tests so I can check how it formats: https://github.com/JuliaRegistries/RegistryCI.jl/blob/34367b742375cb4670bb310b895b0cf33c22fc52/test/reference_comments/comment_pass_false_type_new_package_suggest_onepointzero_true_version_0.1.0_point_to_slack_false.md

Screenshot 2024-12-16 at 22 28 27

IanButterworth:
Good to go from my perspective.

DilumAluthge:
I see. I will remove those two `macos-12` jobs from the list of required status checks.

DilumAluthge:
Disregard. Posted in wrong thread.

PR #590, opened 2024-12-18
Fix guidance if branches

IanButterworth:
These were the wrong way around. (Sorry, should've caught it in my review @ericphanson)


ericphanson:
ah, oops! We may need to regenerate the reference tests: https://github.com/JuliaTesting/ReferenceTests.jl?tab=readme-ov-file#updating-references-within-package-tests

ericphanson (approved):

PR #597, opened 2025-02-20
add a note about pre-v1 breaking changes

IanButterworth:
Ideally we'd not show this if it isn't pre-v1 but I couldn't see a way to do that.

### DIFF
--- a/src/AutoMerge/guidelines.jl
+++ b/src/AutoMerge/guidelines.jl
@@ -325,7 +325,11 @@ end
 # This check checks for an explanation of why a breaking change is breaking
 const guideline_breaking_explanation = Guideline(;
     info = "Release notes have not been provided that explain why this is a breaking change.",
-    docs = "If this is a breaking change, release notes must be given that explain why this is a breaking change (i.e. mention \"breaking\" or \"changelog\"). To update the release notes, please see the \"Providing and updating release notes\" subsection under \"Additional information\" below.",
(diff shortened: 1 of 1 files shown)


DilumAluthge:
> Ideally we'd not show this if it isn't pre-v1 but I couldn't see a way to do that.

We do have this info, so we could figure out a way to propagate it.

ericphanson:
btw this code change only affects the docs, not the actual message being shown to the user in the PR

IanButterworth:
Should be ready for review now.

src/AutoMerge/guidelines.jl
```diff
@@ -332,11 +332,25 @@ function meets_breaking_explanation_check(data::GitHubAutoMergeData)
...
+        This is a breaking change, but the release notes do not mention it. Please add a mention of the breaking change to the release notes (use the words \"breaking\" or \"changelog\").
+        """
+    else
+        """
+        This is a breaking change, but no release notes have been provided. Please add release notes that explain the breaking change.
+        """
+    end
+    if version < v"1.0.0"
+        msg *= """
+        Given this is a pre-v1.0.0 release, you may have not intended to make a breaking change release. [More information](https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0) on Julia's handling of pre-v1.0.0 versioning.
```
ericphanson, line 350: I think we should make this more "actionable", like if this was intended do X, if it was not intended do Y

PR #606, opened 2025-09-16
AutoMerge: Try loading packages for lowest and highest compatible Julia versions

### LINKED ISSUE #602: Try importing package with min and max compatible Julia versions
The check that a package can be imported is today run in a new Julia process using the same Julia version as is running automerge, which would typically be the latest released version. This is a limitation in the following cases:
* Packages/versions which need to restrict Julia compatibility to older versions cannot be automerged.
* Packages/versions which only can run on an upcoming Julia version cannot be automerged.
* Packages might be importable on the current version but have compat of dependencies which makes import fail on other Julia versions, cf. https://discourse.julialang.org/t/pkg-add-customgaussquadrature-error-unsatisfiable-requirements/130147.

Since these days `juliaup` makes it easy to run a variety of Julia versions, it would be better to run the import check on the minimum and maximum Julia versions according to the specified Julia compat. If `juliaup` cannot be assumed to be externally available, the recently registered `juliaup_jll` could be employed.

GunnarFarneback:
~This takes a stab at~ Fixes #602, using `juliaup_jll` to find suitable Julia versions. ~There should be some tests but I'm not sure exactly what and how much we want and can test. It's complete enough to be worth reviewing.~


GunnarFarneback:
A possible workaround could be to have `juliaup_jll` as a weak dependency and activate the improved package loading check as a package extension. I'm not sure it would be a *good* solution but should be doable.

ericphanson:
This version should be unblocked by #610

src/AutoMerge/juliaup.jl
```diff
@@ -0,0 +1,73 @@
+using juliaup_jll: juliaup, julia
+
+const available_julia_versions = VersionNumber[]
+const version_re = r"^(\d+\.\d+\.\d+(-[\w\d]+)?)$"
+
+function find_available_julia_versions()
```
ericphanson, line 6: this looks like it could be a little fragile if `juliaup list` changes its printing, but I don't think there's a better way to do it now. Maybe we could ask for `juliaup list --json` or something to have a stable machine-readable format. But I don't think that needs to block this feature.

ericphanson (dismissed):
implementation looks good to me, had some minor suggestions. Needs to be rebased/merged with latest master obviously, and Julia compat for AutoMerge.jl restricted to v1.6+ (in Project.toml and in CI files).

src/AutoMerge/juliaup.jl
```diff
@@ -0,0 +1,73 @@
...
+
+function get_julia_binary(version, kind)
+    run(`$(juliaup()) add $(version)`)
+    cmd = `$(julia()) +$(version)`
+    text = "julia $(version) ($kind compatible version)"
+    return cmd, text
+end
+
+
+function get_compatible_julia_versions(julia_compat)
```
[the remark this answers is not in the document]
GunnarFarneback: Fair enough. I had planned this function to only return the versions, but it turned out nicer for the accompanying annotations to go straight for the binaries.

GunnarFarneback:
The code to test import does this:
```
    code = """
        import Pkg;
        Pkg.Registry.add(Pkg.RegistrySpec(path=\"$(working_directory)\"));
        _registry_deps = $(_registry_deps);
        _registry_deps_is_valid_url = $(_registry_deps_is_valid_url);
        for i = 1:length(_registry_deps)
            regdep = _registry_deps[i]
            if _registry_deps_is_valid_url[i]
                Pkg.Registry.add(Pkg.RegistrySpec(url = regdep))
            else
                Pkg.Registry.add(regdep)
            end
        end
        @info("Attempting to `Pkg.add` package...");
        $(pkg_add_command)
        @info("Successfully `Pkg.add`ed package");
        @info("Attempting to `import` package");
        Pkg.precompile()
        import $(pkg);
        @info("Successfully `import`ed package");
        """
```
The `Pkg.Registry` calls don't work on 1.0 since they were added in 1.1. The `Pkg.precompile()` call does not work before 1.3.

I don't think we can do much about 1.0 but is it really necessary to do an explicit `Pkg.precompile()` here? I believe it should be done automatically by the `import`.

ericphanson:
We can drop pre-1.6 in AutoMerge.jl (or even pre-1.10 if we wanted). Not sure when the explicit precompile is helpful or not

GunnarFarneback:
If a package claims compat with 1.0 I'd rather test importing on as low version as possible. I'm reasonably certain that `Pkg.precompile()` only is a way to force precompilation to happen immediately rather than when you want to import the package, but since we do it immediately anyway there shouldn't be a difference.

ericphanson:
ah I see, so it's not the version of julia running automerge, it's the version running the subprocess via juliaup. I'm fine with dropping the precompile.

ericphanson:
this LGTM, we just need the required checks to be adjusted by an admin (cc @DilumAluthge)

GunnarFarneback:
Late realization: It doesn't help so much to test importing with carefully chosen Julia versions if it fails anyway on the can-be-added check, which still runs on the same version as AutoMerge itself. So this needs to be extended to adding as well, but it should be enough to run it for the highest compatible version.

ericphanson:
I've been given permission to adjust the required status checks and have done so to match the updated CI config from this PR (i.e. AutoMerge.jl only 1.6, 1.10, 1, while keeping all of them for RegistryCI.jl).

ericphanson (approved):
lgtm, but we should use `try/finally` in the test

ericphanson:
Hah, looks like the integration tests found some lower bound issues already:


```
[ Info: Successfully `Pkg.add`ed package
[ Info: Attempting to `import` package
ERROR: LoadError: Unable to import HelloWorldC_jll on Julia versions older than 1.3!
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] top-level scope at /tmp/jl_S5J0Du/packages/HelloWorldC_jll/TVSQO/src/HelloWorldC_jll.jl:18
 [3] include at ./boot.jl:326 [inlined]
 [4] include_relative(::Module, ::String) at ./loading.jl:1038
 [5] include(::Module, ::String) at ./sysimg.jl:29
 [6] top-level scope at none:2
 [7] eval at ./boot.jl:328 [inlined]
 [8] eval(::Expr) at ./client.jl:404
 [9] top-level scope at ./none:3
in expression starting at /tmp/jl_S5J0Du/packages/HelloWorldC_jll/TVSQO/src/HelloWorldC_jll.jl:3
ERROR: Failed to precompile HelloWorldC_jll [dca1746e-5efc-54fc-8249-22745bc95a49] to /tmp/jl_S5J0Du/compiled/v1.1/HelloWorldC_jll/rt8Uf.ji.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] compilecache(::Base.PkgId, ::String) at ./loading.jl:1197
 [3] _require(::Base.PkgId) at ./loading.jl:960
 [4] require(::Base.PkgId) at ./loading.jl:858
 [5] require(::Module, ::Symbol) at ./loading.jl:853
┌ Error: Was not able to successfully `import` the package on julia 1.1.1 (lowest compatible version)
└ @ AutoMerge ~/work/RegistryCI.jl/RegistryCI.jl/AutoMerge.jl/src/guidelines.jl:1057
┌ Info: Version can be `import`ed
│   meets_this_guideline = false
└   message = "I was not able to load the package on julia 1.1.1 (lowest compatible version)(i.e. `import HelloWorldC_jll` failed). See the AutoMerge logs for details."
Integration tests: Error During Test at /home/runner/work/RegistryCI.jl/RegistryCI.jl/AutoMerge.jl/test/automerge-integration.jl:33
  Got exception outside of a @test
  AutoMerge.AutoMergeGuidelinesNotMet("The automerge guidelines were not met.")
```


https://github.com/JuliaRegistries/RegistryCI.jl/actions/runs/18037755399/job/51328614593

AutoMerge.jl/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "AutoMerge"
 uuid = "c5732277-7834-41e3-8e1f-5f375898cfe1"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>", "contributors"]
-version = "1.0.0"
+version = "1.1.0"
```
ericphanson, line 4: we haven't released v1.0.0 yet so we can include this as part of it

GunnarFarneback:
We need to do something about the failing integration tests, I suppose, but I'm not familiar with how those work.

ericphanson:
I think I did fix it in [ad669f8](https://github.com/JuliaRegistries/RegistryCI.jl/pull/606/commits/ad669f83f2acb371881fd5f7ad288f89a37ee28f), though there may be additional errors

PR #608, opened 2025-09-19
RFC: split out RegistryTesting.jl

ericphanson:
This could allow RegistryCI.jl to increase julia compat without preventing General (or others) from running registry testing on earlier Julia versions. This would be non-breaking for RegistryCI (...at least in normal julia terms, though RegistryCI is often used like an application so upping the julia compat may break workflows) since we can depend on RegistryTesting.jl to provide the `RegistryCI.test` entrypoint still.

context: #606, #607 

used claude-code (for the first time! had tried copilot before but not claude code)

### DIFF
--- a/.github/workflows/ci_unit.yml
+++ b/.github/workflows/ci_unit.yml
@@ -44,6 +44,39 @@ jobs:
         env:
           AUTOMERGE_RUN_INTEGRATION_TESTS: "false"
 
+  registry-testing:
+    name: RegistryTesting/Julia ${{ matrix.version }}/${{ matrix.os }}/${{ matrix.arch }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - x64
+        os:
+          - ubuntu-latest
+        version:
+          - '1.3'
+          - '1.4'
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Yep I would include the TagBotTriggers stuff in AutoMerge.jl.

DilumAluthge:
On a separate note, I've seen people do a top-level `lib/` folder in their repos, and then put subdir packages in there. So maybe we could do `lib/AutoMerge`?

ericphanson:
> On a separate note, I've seen people do a top-level `lib/` folder in their repos, and then put subdir packages in there. So maybe we could do `lib/AutoMerge`?

I think that makes more sense when there's 1 clearly "main" package and then some supplemental ones (and in some cases it seems those are ones that later moved to `ext` with package extensions). Here AutoMerge is bigger (in number of features and amount of code) and I don't think there's a clear "main" package.

DilumAluthge:
Might be nice to keep. That way people (using AutoMerge on private registries) can use AutoMerge as a one-stop shop, with no need to have a separate CI job for the consistency tests.

So AutoMerge.jl would take a direct dep on RegistryCI.jl.

PR #610, opened 2025-09-20
RFC: Split out AutoMerge.jl into its own package as a breaking change

ericphanson:
* RegistryCI v11 only has the registry consistency tests
* AutoMerge v1 depends on RegistryCI, has tagbot stuff too
* General would use RegistryCI on all julia versions, and AutoMerge on latest julia version
* in the future we can remove julia compat from AutoMerge to use latest features/package
* RegistryCI could potentially expand julia compat to 1.0 since it has few deps
     * I noticed it should depend on URIs instead of HTTP but that's an orthogonal fix 
* alternative to #608 

🤖 Generated with [Claude Code](https://claude.ai/code)

I also added some nicer errors:

```julia
julia> RegistryCI.AutoMerge.run()
ERROR: MovedFunctionalityException: RegistryCI.AutoMerge has been moved to its own package, AutoMerge.jl. The API of AutoMerge v1.0 matches that of RegistryCI.AutoMerge v10.10.4.
Stacktrace:
 [1] throw_error(a::RegistryCI.MovedFunctionality)
   @ RegistryCI ~/RegistryCI.jl/src/RegistryCI.jl:22
 [2] getproperty(a::RegistryCI.MovedFunctionality, ::Symbol)
   @ RegistryCI ~/RegistryCI.jl/src/RegistryCI.jl:30
 [3] top-level scope
   @ REPL[12]:1

julia> RegistryCI.TagBot.main()
ERROR: MovedFunctionalityException: RegistryCI.TagBot has been moved to a new package, AutoMerge.jl. The API of AutoMerge.TagBot v1.0 matches that of RegistryCI.TagBot v10.10.4.
Stacktrace:
 [1] throw_error(a::RegistryCI.MovedFunctionality)
   @ RegistryCI ~/RegistryCI.jl/src/RegistryCI.jl:24
 [2] getproperty(a::RegistryCI.MovedFunctionality, ::Symbol)
   @ RegistryCI ~/RegistryCI.jl/src/RegistryCI.jl:30
 [3] top-level scope
   @ REPL[13]:1
```


DilumAluthge:
> RegistryCI could potentially expand julia compat to 1.0 since it has few deps

Yes, we should definitely do this (can be done in follow-up PR).

DilumAluthge:
Currently, in the General registry, we have a single "CI" manifest (or rather, one CI manifest per Julia version).

Let's plan to split that into two: a "CI" manifest (one CI manifest per Julia version), and an "AutoMerge" manifest (just one AutoMerge manifest needed).

ericphanson:
if we go with this, the CI statuses would need to be updated

DilumAluthge:
I think I've fixed the branch protection settings.

Is the Codecov drop expected?

DilumAluthge:
@ericphanson After this PR has been merged, can you make a follow-up PR to add this PR's merge (or squash) commit to the [`.git-blame-ignore-revs` file](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view)?

GunnarFarneback:
If I read this correctly, RegistryCI remains at the top of the repository whereas AutoMerge moves into a subdirectory.

I'd rather see that we do this right and place both packages in subdirectories. Otherwise RegistryCI will be distributed with an unused copy of AutoMerge.

.github/workflows/ci_unit.yml
```diff
@@ -87,6 +138,6 @@ jobs:
       - run: |
           julia --project=docs -e '
             using Documenter: DocMeta, doctest
-            using RegistryCI
-            DocMeta.setdocmeta!(RegistryCI, :DocTestSetup, :(using RegistryCI); recursive=true)
```
DilumAluthge, line 91: Is there a reason we no longer need this?

AutoMerge.jl/test/runtests.jl
```diff
@@ -0,0 +1,54 @@
...
+
+    AUTOMERGE_RUN_INTEGRATION_TESTS =
+        get(ENV, "AUTOMERGE_RUN_INTEGRATION_TESTS", "")::String
+    if AUTOMERGE_RUN_INTEGRATION_TESTS == "true"
+        @testset "AutoMerge.jl integration tests" begin
+            @info("Running the AutoMerge.jl integration tests")
+            include("automerge-integration.jl")
+        end
+    end
+end
```
DilumAluthge, line 54: Can you make sure that all files end in a trailing newline character (for compliance with the POSIX standard)?

AutoMerge.jl/Project.toml
```diff
@@ -0,0 +1,62 @@
...
+
+[compat]
+Base64 = "<0.0.1, 1"
+Dates = "<0.0.1, 1"
+GitHub = "5.2"
+HTTP = "0.8, 0.9.1, 1"
+JSON = "0.19, 0.20, 0.21"
+LibGit2 = "<0.0.1, 1"
+LicenseCheck = "0.2"
+Pkg = "1"
```
DilumAluthge, line 34:
```suggestion
Pkg = "<0.0.1, 1"
```

AutoMerge.jl/Project.toml
```diff
@@ -0,0 +1,62 @@
...
+Printf = "<0.0.1, 1"
+Random = "<0.0.1, 1"
+RegistryTools = "2.2"
+RegistryCI = "10.0, 11.0"
+SHA = "<0.0.1, 0.7, 1"
+StringDistances = "0.9, 0.10, 0.11"
+TOML = "1"
```
DilumAluthge, line 41:
```suggestion
TOML = "<0.0.1, 1"
```

AutoMerge.jl/Project.toml
```diff
@@ -0,0 +1,62 @@
...
+Tar = "1"
```
DilumAluthge, line 42:
```suggestion
Tar = "<0.0.1, 1"
```

docs/make.jl
```diff
@@ -1,10 +1,9 @@
 using RegistryCI
+using AutoMerge
 using Documenter
 
-DocMeta.setdocmeta!(RegistryCI, :DocTestSetup, :(using RegistryCI); recursive=true)
-
```
DilumAluthge, line 5: Similar comment to above - why do we no longer need this?

AutoMerge.jl/Project.toml
```diff
@@ -0,0 +1,62 @@
...
+GitHub = "5.2"
+HTTP = "0.8, 0.9.1, 1"
+JSON = "0.19, 0.20, 0.21"
+LibGit2 = "<0.0.1, 1"
+LicenseCheck = "0.2"
+Pkg = "1"
+Printf = "<0.0.1, 1"
+Random = "<0.0.1, 1"
+RegistryTools = "2.2"
+RegistryCI = "10.0, 11.0"
```
DilumAluthge, line 38: Could be simpler as this?

```suggestion
RegistryCI = "10, 11"
```

GunnarFarneback:
It's a little hard to tell but I don't think I can see a license inside the AutoMerge package.

src/RegistryCI.jl
```diff
@@ -2,9 +2,31 @@ module RegistryCI
...
+const AutoMerge = MovedFunctionality("AutoMerge")
+const TagBot = MovedFunctionality("TagBot")
+function throw_error(a::MovedFunctionality)
+    name = getfield(a, :name)
+    if name == "AutoMerge"
+        throw(MovedFunctionalityException("RegistryCI.AutoMerge has been moved to its own package, AutoMerge.jl. The API of AutoMerge v1.0 matches that of RegistryCI.AutoMerge v10.10.4."))
+    elseif name == "TagBot"
+        throw(MovedFunctionalityException("RegistryCI.TagBot has been moved to a new package, AutoMerge.jl. The API of AutoMerge.TagBot v1.0 matches that of RegistryCI.TagBot v10.10.4."))
+    else
+        @assert false
```
DilumAluthge, line 26: We may disable `@assert` in the future (depending on optimization level).
```suggestion
        error("Invalid value for name: $name")
```

DilumAluthge:
> If I read this correctly, RegistryCI remains at the top of the repository whereas AutoMerge moves into a subdirectory.
> 
> I'd rather see that we do this right and place both packages in subdirectories. Otherwise RegistryCI will be distributed with an unused copy of AutoMerge.

I'm fine with putting both packages in subdirectories.

.github/workflows/ci_unit.yml
ericphanson: it wasn't used on main either; I removed it since I got confused how to update it, then realized it wasn't needed

ericphanson:
thanks for the review folks, I think I've addressed them:

* RegistryCI.jl moved to subdir
* LICENSES added
* Project formatting suggestions applied

DilumAluthge (dismissed):
The CI badges in the README look broken. That can be fixed in a follow-up PR.

Otherwise, LGTM.

DilumAluthge:
@ericphanson Any idea why the Codecov is dropping?

ericphanson:
it says 2 lines are missing:

Screenshot 2025-09-21 at 00 02 34

One case is supposed to never be hit (the name thing), the other I did add a test for

ericphanson:
I think that's since integration tests haven't run on the feature branch but they have run on master

DilumAluthge (dismissed):

DilumAluthge:
This is good to merge from my POV, once Gunnar has finished his review.

GunnarFarneback (dismissed):

GunnarFarneback:
I guess I didn't review this carefully enough.
```
├── AutoMerge.jl
├── bin
├── bors.toml
├── codecov.yml
├── docs
├── example_github_workflow_files
├── INTEGRATION_TESTS.md
├── LICENSE.md
├── Manifest.toml
├── README.md
└── RegistryCI.jl
```

I can't say I'm a fan of the `.jl` suffix for the package subdirectories.
* From a technical point of view these directories can be named anything.
* The *repository* naming convention for packages doesn't really apply; it's mostly a convention for making it easier to search for those repositories.
* Inside the repository you would expect something ending with `.jl` to be a Julia source file. This becomes particularly confusing in a view that doesn't clearly mark directories from files.

ericphanson:
Hm, we can rename it if you want. I am quite used to this convention and use it in mono-repos at work. I have found it helpful to distinguish between julia packages and other directories. But I don't have that strong of an opinion here, and am fine with it changing. We haven't registered but if we are going to change it we should do so before registration to reduce churn, IMO.

DilumAluthge:
I don't feel strongly either way.

PR #611, opened 2025-09-21
fix unit tests: don't test SnoopCompileCore's repo

ericphanson:
since it has moved and we don't control it. The repo field is tested earlier anyway. Cherry-picked out of #609 since CI is also failing on #610 due to this. IIUC we only test SnoopCompileCore since it is an example of a subdir package.

This is included in #610 so if that gets through the merge queue we can just close this.


ericphanson:
Closed by #610

PR #613, opened 2025-09-24
Move configuration to struct and add `GENERAL_AUTOMERGE_CONFIG`

### LINKED ISSUE #612: Proposal: move General's configuration of AutoMerge from General repo into AutoMerge.jl itself
## Proposal

I propose adding a configuration struct

```julia
Base.@kwdef struct AutoMergeConfiguration
   merge_new_packages::Bool
   merge_new_versions::Bool
   new_package_waiting_period::Dates.TimePeriod
   ...
end
```

with all the kwargs from `run`:

https://github.com/JuliaRegistries/RegistryCI.jl/blob/f4573cdd684e218a82e3339ff63fa8d2c7d64d71/AutoMerge.jl/src/public.jl#L14-L40

We would document that adding new fields with defaults to this struct is non-breaking, but removing or renaming existing fields is breaking.

Then we would make
```julia
const GENERAL_CONFIG = AutoMergeConfiguration(; ...bunch of stuff...)
```

as a constant in the package. We would document that this config is allowed to change in non-breaking releases, following decisions by the maintainers of General to change parameters.

We would refactor run to have the signature

```julia
run(config::AutoMergeConfiguration, env=ENV, ciccfg::CIService=auto_detect_ci_service(; env=env); config_overrides...)
```

in which the kwargs `config_overrides` serve to override values of `config`. Thus, General will invoke

```julia
run(AutoMerge.GENERAL_CONFIG)
```
in typical use, and

```julia
run(AutoMerge.GENERAL_CONFIG; read_only=true)
```

when it's running a  "staging" version of AutoMerge.

Likewise private registries can do

```julia
run(AutoMerge.GENERAL_CONFIG; new_package_waiting_period=Minute(45))
```

to say "use the same config as General, except with a shorter waiting period", for example, if they want to opt-in to following General's semantics. We may want to discourage this though if they don't want to be surprised by changes in non-breaking releases, and suggest they should specify their own config.

## Why?

1. my motivating reason is to support #453: if we add support to AutoMerge.jl to run locally on a package, we want settings like `check_license`, `check_breaking_explanation`, `public_registries`, etc to match General without users having to look up what those settings are. If we don't codify General's choices in AutoMerge, then a generic `check_local(...)` would easily get out-of-sync with what settings General actually uses.
2. This makes the YAMLs smaller in General, with less to copy-paste-modify for private registries
3. Grouping kwargs into a struct is generally a nice pattern

ericphanson:
closes #612 

Once we release AutoMerge.jl we can update General to use it, following the examples here.

### DIFF
--- a/AutoMerge.jl/src/types.jl
+++ b/AutoMerge.jl/src/types.jl
@@ -1,6 +1,115 @@
+using Dates: Day, Minute
+
 struct AlwaysAssertionError <: Exception
     msg::String
 end
+
+
+"""
+    AutoMergeConfiguration
+
+Configuration struct for AutoMerge.
+
+```julia
+AutoMergeConfiguration(; kwargs...)
+```
+
+!!! note
+    New keyword arguments with defaults may be added to this struct in _non-breaking_ releases of AutoMerge.jl. Default values and keyword argument names will not be removed or changed in non-breaking releases, however.
+
+# Required keyword arguments
+
+- `merge_new_packages::Bool`: should AutoMerge merge registration PRs for new packages
+- `merge_new_versions::Bool`: should AutoMerge merge registration PRs for new versions of packages
+- `new_package_waiting_period::Dates.Period`: new package waiting period, e.g `Day(3)`.
+- `new_jll_package_waiting_period::Dates.Period`: new JLL package waiting period, e.g `Minute(20)`.
+- `new_version_waiting_period::Dates.Period`: new package version waiting period, e.g `Minute(10)`.
+- `new_jll_version_waiting_period::Dates.Period`: new JLL package version waiting period, e.g `Minute(10)`.
+- `registry::String`: the registry name you want to run AutoMerge on.
+- `authorized_authors::Vector{String}`: list of who can submit registration, e.g `String["JuliaRegistrator"]`.
+- `authorized_authors_special_jll_exceptions::Vector{String}`: a list of users who can submit JLL packages (which have strict rules about allowed dependencies and are subject to `new_jll_*_waiting_period`s instead of `new_*_waiting_period`s).
+
+# Keyword arguments with default values
+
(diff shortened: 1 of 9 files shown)


AutoMerge.jl/src/types.jl
```diff
@@ -1,7 +1,137 @@
...
+    print(io, AutoMergeConfiguration, " with:")
+    for k in propertynames(obj)
+        print(io, "\n  ", k, ": `", repr(getproperty(obj, k)), "`")
+    end
+end
+
+Base.show(io::IO, ::AutoMergeConfiguration) = print(io, AutoMergeConfiguration, "(…)")
+
+
+const GENERAL_AUTOMERGE_CONFIG = AutoMergeConfiguration(
```
ericphanson, line 91: values match https://github.com/JuliaRegistries/General/blob/3fa5003de7b8fa925c7d0d5fa783cafe07b3830a/.github/workflows/automerge.yml#L98-L117 except
```
merge_new_packages = true,
 merge_new_versions = true
```

are unconditional instead of based on ENV variables. We can override it with ENVs if needed in General.

docs/src/private-registries.md
```diff
@@ -57,27 +57,14 @@ In order to enable automerge support, you will also have to copy the `automerge.
 
 ```julia
 using AutoMerge
+using AutoMerge: AutoMergeConfiguration
 using Dates
-AutoMerge.run(
```
ericphanson, line 61: one less thing to keep in sync now

ericphanson:
CI on pre-1.5 will fail; we remove support in https://github.com/JuliaRegistries/RegistryCI.jl/pull/606 so I'll wait for that

ericphanson:
this config gets split apart and refactored in https://github.com/JuliaRegistries/RegistryCI.jl/pull/617, so I'm not sure it's worth reviewing this PR

Issue #453, opened 2022-08-03
Run AutoMerge checks in package


GunnarFarneback:
AutoMerge is run on PRs to the General registry in order to determine whether they can be merged automatically (well, yeah, that's what the name means).

However, failures to pass automerge typically requires making package updates, a new commit, and trigger a new registration. This is a relatively slow feedback loop for those automerge checks which can be performed locally by the package. It would be nice if packages could opt in to run applicable checks in their own CI tests by adding something similar to
```
using RegistryCI
check_if_my_package_is_good_for_automerge()
```
to their `runtests.jl`.

ericphanson:
How should it handle the distinction between new packages and new versions? Simplest would be a keyword argument, but it also could try to inspect the registry or something to see if it is already registered (I guess via RegistryInstances)

ericphanson:
> After [#606](https://github.com/JuliaRegistries/RegistryCI.jl/pull/606) I think we definitely want to run the import checks in local testing, otherwise it's easy to miss an incorrect lower bound on Julia.

Good point 

> It's questionable whether we want to run checks like `guideline_registry_consistency_tests_pass`. Those have nothing to do with the package and in case they should happen to fail, there's nothing the package author can do about it.

True, we can probably drop that

> There is a way to avoid having to copy the General registry. We could create a new registry containing only the new package/version and rely on Pkg merging the registry information. Some of the code might need adjustments, but I don't think it should be that much.

I don’t think this works in the current setup: we are very strict and use a new depot with only the PR-registry installed to simulate what would happen if the PR were merged to be the new version of General. So there is no old version of General to merge with.

DilumAluthge:
> Side note: All file inspection done in AutoMerge could be done on registry tarballs with the help of the CodecZLib and Tar packages.

But in prod, we don't have a tarball, right? We have a locally checked out Git clone.

GunnarFarneback:
Still have to pay the cost for many small files, when all we actually need is a half dozen files or so.

RegistryInstances can simplify some of the AutoMerge code but won't help if we want to run RegistryTools.

GunnarFarneback:
For current use: no functional changes.

File reading: Abstract it out to a function which can read normal files as usual and from a tarball via CodecZLib and Tar.

[...]

PR #614, opened 2025-09-25
AutoMerge: Refactor parsing of registry toml files.

GunnarFarneback:
Prepares for https://github.com/JuliaRegistries/RegistryCI.jl/issues/453#issuecomment-3330925354 but is a good idea regardless.

### DIFF
--- a/AutoMerge.jl/src/toml.jl
+++ b/AutoMerge.jl/src/toml.jl
@@ -1,16 +1,6 @@
-const _toml_cache = Dict{String, Any}()
-
-# Cache parsed toml files for efficiency.
-# Important: Do not mutate the returned data, as it will corrupt the cache.
 function parse_registry_toml(registry_dir, path_components...; allow_missing = false)
     path = joinpath(registry_dir, path_components...)
-    get!(_toml_cache, path) do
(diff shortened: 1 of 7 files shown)


AutoMerge.jl/src/toml.jl
```diff
@@ -1 +1,16 @@
-maybe_parse_toml(f::AbstractString) = ispath(f) ? Pkg.TOML.parsefile(f) : Dict{String, Any}()
+const _toml_cache = Dict{String, Any}()
+
+# Cache parsed toml files for efficiency.
+# Important: Do not mutate the returned data, as it will corrupt the cache.
```
GunnarFarneback, line 4: This is a downside. Alternatives are not to cache or to return a deepcopy of the cached value.

ericphanson (commented):
hm, do we actually parse the files many times in a single run? I'm not really against the cache but in my mental model it doesn't seem super necessary. I guess the main idea is to centralize access so we can provide alternate ways to read the files?

GunnarFarneback:
I backed out the caching. It can easily be re-added later if it seems useful.

PR #629, opened 2025-09-29
AutoMerge: improve diff styling and add diff stats

ericphanson:
Followup to https://github.com/JuliaRegistries/RegistryCI.jl/pull/609

This PR adds the output of

```sh
git -C $clone_dir diff-tree --stat $old_tree_sha $new_tree_sha --stat-count=10 --stat-width=80 --no-color
```

to the AutoMerge comment, in addition to the link. We can limit the output to 10 lines (with `--stat-count`) and we can run this even for non-github hosted repos.

Maybe it's too verbose? We could put it in a `<details>` block. Or we could use `--shortstat` instead, example below.

## Example with `--stat --stat-count=10 --stat-width=80` 

Screenshot 2025-09-30 at 01 17 10

## Example with `--shortstat` instead

Screenshot 2025-09-30 at 01 22 47


AutoMerge/test/format-diff-stats.jl
```diff
@@ -0,0 +1,193 @@
...
+         props = TestProps(has_utf8_msg=true, has_shortstat=true)),
+    ]
+
+    @testset "$(tc.name)" for tc in test_cases
+        result = AutoMerge.format_diff_stats(tc.full_diff, tc.stat, tc.shortstat;
+                                             old_tree_sha=old_sha, new_tree_sha=new_sha)
+        if length(result) < 1000
+        println("--------$(tc.name)--------")
+        println(result)
+        end
```
ericphanson, line 188: oops, some manual checking I don't want to commit
```suggestion
```

PR #643, opened 2025-10-22
AutoMerge: switch to RegistryInstances

ericphanson:
this ends up kind of undoing #614 because it seemed easier to use RegistryInstances everywhere. Now we can use packed registries everywhere except the RegistryCI.jl-based check which needs to support older Julia's. In particular, this should make it more feasible to set up a "local" version that users can use without needing a unpacked registry.

written by claude-code but with a very annoying amount of iteration, probably it would have been faster to do it myself.

The most gnarly checks here are the deps and julia compat, and I tested that we got the same results on the latest release of all 11k packages (this did reveal a bug in the initial implementation, in which there was only a julia dep and no other deps). I also included some tests with the old version on a few packages to show they are the same, but we could delete those tests before merging if we want.


DilumAluthge:
So, if I understand correctly, you're only using RegistryInstances (and the packed registry) for the upstream registry head, right? For the checked-out PR branch, you're still using the unpacked registry that we have checked out on disk?

Do I have that correct?

ericphanson:
@GunnarFarneback I know you might not have time for a full review, but do you have any thoughts on this approach?

ericphanson:
IIRC this is good to go, but I don't have time to revive it now, feel free to take over

PR #653, opened 2026-02-22
Update the TagBot workflow for the two subdir packages (AutoMerge, RegistryCI)

DilumAluthge:


GunnarFarneback:
Sorry, I have little knowledge about actions and none about TagBot.

PR #657, opened 2026-03-02
AutoMerge: Require version numbers to be `≥ 0.1.0` (except for Yggdrasil JLLs)

DilumAluthge:
Continue to accept `0.1.0`, `1.0.0`, and `X.0.0`

### DIFF
--- a/AutoMerge/src/guidelines.jl
+++ b/AutoMerge/src/guidelines.jl
@@ -830,7 +830,7 @@ const guideline_no_zero_dot_zero_version_number = Guideline(;
 )
 
 function meets_no_zero_dot_zero_version_number(version)
-    meets_this_guideline = version.major != 0
+    meets_this_guideline = version >= "0.1.0"
     if meets_this_guideline
         return true, ""
     else
--- a/AutoMerge/src/guidelines.jl
+++ b/AutoMerge/src/guidelines.jl
@@ -824,27 +824,18 @@ function meets_sequential_version_number(
     return meets_sequential_version_number(_all_versions, new_version)
 end
 
-const guideline_standard_initial_version_number = Guideline(;
-    info="Standard initial version number. Must be one of: `0.0.1`, `0.1.0`, `1.0.0`, or `X.0.0`.",
-    check=data -> meets_standard_initial_version_number(data.version),
+const guideline_no_zero_dot_zero_version_number = Guideline(;
+    info="Version number must not be of the form 0.0.x",
+    check=data -> meets_no_zero_dot_zero_version_number(data.version),
 )
 
-function meets_standard_initial_version_number(version)
-    meets_this_guideline =
-        version == v"0.0.1" ||
-        version == v"0.1.0" ||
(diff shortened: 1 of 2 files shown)


goerz (commented):
Does this do anything for the actual `General` registry? I don't think the `meets_standard_initial_version_number` is active: It's not listed in `get_automerge_guidelines(::NewPackage)` further down in the file. And I've never seen this particular check applied. We pretty commonly have `0.x.0` initial version for `x > 1`, and the only effect I've seen is the registrator bot recommending to use a "standard" initial version, in response to a `@JuliaRegistrator register` comment (i.e., not in the registration PR in `General`, but in the original project repo) .

I would speculate that `meets_standard_initial_version_number` is either left over and can be deleted, or is intended for other registries that may want to impose this rule. If that's the case, then of course it does make sense to remove `0.0.1` here; but it seems to me that for `General`, we would need a separate (new) `Guideline` for ```"Initial version number must be >= `0.1.0`"``` that will have to be added to the list of `guidelines` inside `get_automerge_guidelines(::NewPackage)` further down.

DilumAluthge:
Ah. You are 100% correct. We disabled it in https://github.com/JuliaRegistries/RegistryCI.jl/pull/275

@ericphanson Have we tagged AutoMerge v1.0 yet? If not, what do you think of making the following breaking change as part of AutoMerge v1.0:

1. Modify this guideline to implement Michael's proposal, namely the guideline checks (for both new packages and new versions) that the version is >= 0.1.0
2. Rename this guideline to reflect its new behavior
3. Add the guideline to both new packages and new versions

DilumAluthge:
> > and the only effect I've seen is the registrator bot recommending to use a "standard" initial version
> 
> Yes, this is a warning from Registrator, which can be found in RegistryTools: https://github.com/JuliaRegistries/RegistryTools.jl/blob/master/src/register.jl#L77-L80

https://github.com/JuliaRegistries/RegistryTools.jl/pull/110

AutoMerge/src/guidelines.jl
```diff
@@ -824,29 +824,20 @@ function meets_sequential_version_number(
...
-        version == v"0.1.0" ||
-        version == v"1.0.0" ||
-        _is_x_0_0(version)
+function meets_no_zero_dot_zero_version_number(version)
+    meets_this_guideline = version >= "0.1.0"
     if meets_this_guideline
         return true, ""
     else
-        return false, "Version number is not 0.0.1, 0.1.0, 1.0.0, or X.0.0"
+        return false, "Version number must not be of the form 0.0.x"
```
goerz:
```
return false, "Version number must be ≥ 0.1.0",
```

goerz (changes requested):
LGTM, except that maybe "Version number must not be of the form 0.0.x" isn't the most straightforward formulation of the error message, and I would recommend simply saying "Version number must be ≥ 0.1.0"

goerz (dismissed):

PR #671, opened 2026-03-20
RegistryCI: require package repo URLs to end in .git (except for repos hosted on Sourcehut)

### LINKED ISSUE #556: Consistency CI doesn't check if package link ends in `.git`
When doing this migration: https://github.com/JuliaRegistries/General/pull/100032 I accidentally dropped the `.git` from the package URL, and somehow the consistency CI didn't catch it. (fixed here: https://github.com/JuliaRegistries/General/pull/100034)

It'd be good if we had a check for that in the CI.

DilumAluthge-LLM:
Fixes #556.

## Summary

- enforce that each package `repo` entry in `Package.toml` ends with `.git`
- skip that requirement for repositories hosted on Sourcehut (`git.sr.ht`)
- keep the check in the existing registry consistency test path so malformed package links fail validation early

cc @DilumAluthge

🤖 Generated by Codex.

### DIFF
--- a/RegistryCI/src/registry_testing.jl
+++ b/RegistryCI/src/registry_testing.jl
@@ -170,6 +170,10 @@ function test(path=pwd(); registry_deps::Vector{<:AbstractString}=String[])
                 @test Base.isidentifier(data["name"])
                 @test haskey(pkg, "repo")
                 if !is_sourcehut_repo(pkg["repo"])
+                    # Sourcehut does not seem to support the URL form that ends in `.git`
+                    # So we have to skip Sourcehut repos for this check
+                    #
+                    # For all non-Sourcehut repo, we require that the repo URL ends in `.git`
                     @test endswith(pkg["repo"], ".git")
                 end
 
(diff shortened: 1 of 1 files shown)


DilumAluthge:
I have manually reviewed the code in this PR.

DilumAluthge:
@ericphanson @GunnarFarneback Do you think this would be a breaking change for folks with custom registries? Should we move this to AutoMerge instead? (Right now, it's in the RegistryCI registry consistency tests.)

GunnarFarneback:
Yes, I believe that there exist registries where this would be both breaking and unwanted. It's also more a matter of policy than a technical requirement on a sound and working registry.

On the other hand I'm not sure it would be very helpful to do this check in automerge since the most interesting use case is to check manual  URL changes.

DilumAluthge:
Perhaps I should put this check inside of Treecheck? Which will run on repo URL changes.

GunnarFarneback:
It doesn't fit the name of the check so well, but I agree that it runs in the right place.

PR #716, opened 2026-05-05
AutoMerge: allow configuring GitHub Actions username

### LINKED ISSUE #714: `username` hardcodes the result
https://github.com/JuliaRegistries/RegistryCI.jl/blob/c4aefc0d3f3a9f59b5b50812a00985056c21cec4/AutoMerge/src/ciservice.jl#L83-L87 this is obviously wrong.  First of all, this shows utterly confusing
```
[ Info: Authenticated to GitHub as "github-actions[bot]"
```
messages also when using other tokens
https://github.com/JuliaRegistries/RegistryCI.jl/blob/c4aefc0d3f3a9f59b5b50812a00985056c21cec4/AutoMerge/src/public.jl#L63
Also, as an additional side effect, if you use a non `github-actions[bot]` token then the AutoMerge message is never updated but always posted once more, because the code to update the message https://github.com/JuliaRegistries/RegistryCI.jl/blob/c4aefc0d3f3a9f59b5b50812a00985056c21cec4/AutoMerge/src/automerge_comment.jl#L19-L21 searches for comments by `whoami`, but that's hardcoded to `github-actions[bot]`, which, again, is bad.

DilumAluthge-LLM:
## Summary

Fixes #714.

This makes the GitHub Actions AutoMerge username configurable through the existing public `cicfg` argument. Callers can now pass `AutoMerge.GitHubActions(; username="my-app[bot]")` to `check_pr` or `merge_prs` when using a custom token or GitHub App token whose comment author is not `github-actions[bot]`.

The default remains `github-actions[bot]`, so existing callers that rely on auto-detected GitHub Actions behavior are unchanged.

## Validation

- `git diff --check`
- CI will run the test suite for this PR.

CC: @DilumAluthge

🤖 Generated by Codex.

### DIFF
--- a/AutoMerge/src/ciservice.jl
+++ b/AutoMerge/src/ciservice.jl
@@ -39,11 +39,9 @@ end
 ####################
 ## GitHub Actions ##
 ####################
-struct GitHubActions <: CIService
-    enable_cron_builds::Bool
-    function GitHubActions(; enable_cron_builds=true)
-        return new(enable_cron_builds)
-    end
+Base.@kwdef struct GitHubActions <: CIService
+    enable_cron_builds::Bool = true
+    username::String = "github-actions[bot]"
 end
 
 function conditions_met_for_pr_build(cfg::GitHubActions; env=ENV, kwargs...)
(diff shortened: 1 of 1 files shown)


AutoMerge/src/ciservice.jl
```diff
@@ -80,10 +80,10 @@ end
 function directory_of_cloned_registry(cfg::GitHubActions; env=ENV, kwargs...)
     return get(env, "GITHUB_WORKSPACE", nothing)
 end
-function username(api::GitHub.GitHubAPI, cfg::GitHubActions; auth)
-    # /user endpoint of GitHub API not available
-    # with the GITHUB_TOKEN authentication
-    return "github-actions[bot]"
+function username(
+    api::GitHub.GitHubAPI, cfg::GitHubActions; auth, username="github-actions[bot]"
```
DilumAluthge: Don't have default value for `username`.

DilumAluthge: `api::GitHub.GitHubAPI` --> `::GitHub.GitHubAPI`

`cfg::GitHubActions` --> `::GitHubActions`

Make this function one-line.

DilumAluthge:
I think this is the easiest (and most minimal) way to fix #714.

PR #719, opened 2026-06-16
AutoMerge: Add an App to locally check naming guidelines

GunnarFarneback:
It's about time AutoMerge gets an app.

```
$ general_name_check csv
✔ Name is a valid identifier.
✔ Name only contains ascii characters.
✖ Name matches registered package up to capitalization:
  Package name matches existing package name CSV up-to-case.
✖ Name is shorter than five characters:
  Name is not at least 5 characters long
✔ Name passes julia redundancy check.
✖ Name does not have normal capitalization:
  Name does not meet all of the following: starts with an upper-case letter, ASCII alphanumerics only, not all letters are upper-case.
✖ Name similarities found:
  Package name similar to 3 existing packages.
    1. Similar to Rsvg. Damerau-Levenshtein distance 2 is at or below cutoff of 2.
    2. Similar to Vcov. Damerau-Levenshtein distance 2 is at or below cutoff of 2.
    3. Similar to uCSV. Damerau-Levenshtein distance 1 between lowercased names is at or below cutoff of 1.

$ general_name_check CommaSeparatedValues
✔ Name is a valid identifier.
✔ Name only contains ascii characters.
✔ Name is not already registered.
✔ Name is at least five characters.
✔ Name passes julia redundancy check.
✔ Name has normal capitalization.
✔ No name similarity found.
```

This PR has been made without AI support and should be fairly easy to review.

To try this out before it's merged, install the app with
```
using Pkg
Pkg.Apps.add(url="https://github.com/JuliaRegistries/RegistryCI.jl.git", subdir="AutoMerge", rev="name_check_app")
```

### DIFF
--- a/AutoMerge/Project.toml
+++ b/AutoMerge/Project.toml
@@ -27,9 +27,6 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 VisualStringDistances = "089bb0c6-1854-47b9-96f7-327dbbe09dca"
 juliaup_jll = "038fdbe3-3777-5af9-9568-3738956aeb97"
 
-[sources]
-RegistryCI = {path = "../RegistryCI"}
-
 [compat]
 Base64 = "<0.0.1, 1"
 Dates = "<0.0.1, 1"
--- a/AutoMerge/Project.toml
+++ b/AutoMerge/Project.toml
@@ -1,7 +1,7 @@
 name = "AutoMerge"
 uuid = "c5732277-7834-41e3-8e1f-5f375898cfe1"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Fredrik Ekre <ekrefredrik@gmail.com>", "contributors"]
-version = "1.0.0"
+version = "1.1.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
@@ -15,6 +15,7 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RegistryCI = "0c95cc5f-2f7e-43fe-82dd-79dbcba86b32"
+RegistryInstances = "2792f1a3-b283-48e8-9a74-f99dce5104f3"
 RegistryTools = "d1eb7eb1-105f-429d-abf5-b0f65cb9e2c4"
 SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
 StringDistances = "88034a9c-02f8-509d-84a9-84ec65e18404"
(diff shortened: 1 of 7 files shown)


AutoMerge/Project.toml
```diff
@@ -26,9 +27,6 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 VisualStringDistances = "089bb0c6-1854-47b9-96f7-327dbbe09dca"
 juliaup_jll = "038fdbe3-3777-5af9-9568-3738956aeb97"
 
-[sources]
-RegistryCI = {path = "../RegistryCI"}
-
```
GunnarFarneback, line 31: This section was removed to work around a Pkg app installation bug. Do we actually need it?

DilumAluthge: IIRC, I believe we need this to make CI work correctly. What's the Pkg bug?

GunnarFarneback: Before the commit removing `[sources]`:
```
(@v1.12) pkg> app add AutoMerge#b25a6d53fce9054d724c90547c7195fee54d64bf
[ Info: Use `./AutoMerge` to add or develop the local directory at `~/.julia/dev/RegistryCI/AutoMerge`.
    Updating registry at `~/.julia/registries/General`
┌ Info: The General registry is installed via git. Consider reinstalling it via
│ the newer faster direct from tarball format by running:
│   pkg> registry rm General; registry add General
└ 
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
    Updating registry at `~/.julia/registries/Local`
    Updating git-repo `file:///home/gunnar/.julia/registries/Local`
ERROR: expected package `RegistryCI [0c95cc5f]` to exist at path `/home/gunnar/.julia/environments/apps/RegistryCI`
```
Also see recent discussion in slack Pkg-dev.

If it affects CI that's presumably to get master RegistryCI when testing master AutoMerge. It's not obvious to me why we need that but if it is, could we get the same effect with a repo top-level `Project.toml` which sources both RegistryCI and AutoMerge?

AutoMerge/src/general_name_check.jl
```diff
@@ -0,0 +1,90 @@
+module GeneralNameCheck
```
DilumAluthge, line 1: To clarify that this module is only used in the app?

```suggestion
module GeneralNameCheckApp
```

ericphanson (approved):
this seems nice and simple. Awhile ago I started working on something similar but I wanted to cover all checks which meant we needed the package source, not just the name, so needed a registry branch w/ the repo (https://github.com/JuliaRegistries/RegistryCI.jl/compare/master...eph/local) which was a lot more complicated.

code looks good but not sure about removing [sources], I think we needed that for something

README.md
```diff
@@ -29,3 +29,23 @@ This repository contains two Julia packages:
 Starting with RegistryCI v11.0, the automerge and TagBot functionality has been moved to the separate AutoMerge.jl package. RegistryCI.jl now focuses solely on registry testing.
 
 Please see the [documentation](https://JuliaRegistries.github.io/RegistryCI.jl/stable) for both packages.
+
+## Locally Check if a Package Name Passes Automerge Name Checks.
```
ericphanson, line 33: should we add something to the hosted docs also?

AutoMerge/src/general_name_check.jl
GunnarFarneback: Sure, I'll get to it tomorrow. (It needs to be updated in at least one more place.)

README.md
GunnarFarneback: No idea how it fits in there.

PR #722, opened 2026-06-24
AutoMerge: Allow subdir change in new version

GunnarFarneback:
If someone moves a package to a subdir and then registers a new version, AutoMerge will be upset because there is a change to `Package.toml` in addition to the normally changed registry files. To resolve this we ask the author to make a manual PR for the addition of the subdir only. This works but is a bit silly since it adds

* A strange automerge error that the package author might not understand.
* Extra work for the package author to make a manual PR.
* Extra work for registry maintainers to review and merge that PR.

There is not that much that can go wrong when moving to a subdir (unless you do it with a git history rewrite) and these days we have a treecheck job that runs when `Package.toml` changes.

Hence this PR relaxes the check of changed files for a new version to also allow changes in `Package.toml`, conditional on only adding, removing, or modifying a subdir line.

This also includes simplifying the `pr_only_changes_allowed_files` logic:
* Skip the number of changed files shortcut. This only rarely triggers, and probably never after these changes, and provides obscure diagnostics.
* Skip the consistency check of number of files from different GitHub APIs.
* Don't do a deep if/else nesting when we anyway explicitly return from every branch.
* Get rid of the `g0` and `m0` variable names. What do they mean anyway?

### DIFF
--- a/AutoMerge/src/changed_files.jl
+++ b/AutoMerge/src/changed_files.jl
@@ -12,6 +12,8 @@ function allowed_changed_files(::NewPackage, pkg::String)
     return result
 end
 
+# Package.toml is only allowed to change subdir information, which is
+# checked separately.
 function allowed_changed_files(::NewVersion, pkg::String)
     _package_relpath_per_scheme = _get_package_relpath_per_name_scheme(; package_name=pkg)
     result = String[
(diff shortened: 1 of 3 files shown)


DilumAluthge:
Can you split this up into two PRs? One NFC PR that does refactoring/simplifying, and a second PR that implements the behavior change?

GunnarFarneback:
Sure, #723.

PR #724, opened 2026-06-26
AutoMerge: recognize files named 'tagbot.yml' as tagbot files

JamesWrigley:
Currently it only checks for the presence of certain strings in the file, which will not trigger for SciML's new standard tagbot file, e.g.: https://github.com/SciML/CurveFit.jl/blob/master/.github/workflows/TagBot.yml

Checking for 'tagbot.yml' (any case) should be safe since that's the name recommended by the TagBot readme. @ChrisRackauckas, I believe this is why tagbot wasn't running for the recent CurveFit releases.

Debugged and written with Codex :robot:

### DIFF
--- a/AutoMerge/test/tagbot-unit.jl
+++ b/AutoMerge/test/tagbot-unit.jl
@@ -1,3 +1,4 @@
+using Base64: base64encode
 using BrokenRecord: BrokenRecord, HTTP, playback
 using Dates: DateTime, Day, UTC, now
 using AutoMerge: TagBot
@@ -49,6 +50,17 @@ end
         @test occursin("JuliaRegistries/TagBot", contents)
         @test TB.tagbot_file("JuliaWeb/HTTP.jl") !== nothing
         @test TB.tagbot_file("JuliaWeb/HTTP.jl"; issue_comments=true) === nothing
+    end
+    tagbot_yml = (; typ="file", path=".github/workflows/TagBot.yml")
+    unrelated_yml = (; typ="file", path=".github/workflows/CI.yml")
+    mock(
+        GH.directory => Mock(([tagbot_yml, unrelated_yml], Dict())),
+        GH.file => Mock((; content=base64encode("uses: SciML/.github/.github/workflows/tagbot.yml@v1"))),
+    ) do directory, file
+        path, contents = TB.tagbot_file("SciML/CurveFit.jl")
+        @test path == ".github/workflows/TagBot.yml"
+        @test occursin("SciML/.github/.github/workflows/tagbot.yml", contents)
+        @test TB.tagbot_file("SciML/CurveFit.jl"; issue_comments=true) === nothing
     end
 end
 
--- a/AutoMerge/src/TagBot/TagBot.jl
+++ b/AutoMerge/src/TagBot/TagBot.jl
@@ -73,7 +73,9 @@ function tagbot_file(repo; issue_comments=false)
         f.typ == "file" || continue
         file = GH.file(repo, f.path; auth=AUTH[])
         contents = String(base64decode(file.content))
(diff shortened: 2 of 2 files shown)


JamesWrigley:
Nightly failure is a segfault from libcurl, which I believe is unrelated.

JamesWrigley:
(bump)

JamesWrigley:
(bump)

AutoMerge/test/tagbot-unit.jl
```diff
@@ -50,6 +51,17 @@ end
         @test TB.tagbot_file("JuliaWeb/HTTP.jl") !== nothing
         @test TB.tagbot_file("JuliaWeb/HTTP.jl"; issue_comments=true) === nothing
     end
+    tagbot_yml = (; typ="file", path=".github/workflows/TagBot.yml")
+    unrelated_yml = (; typ="file", path=".github/workflows/CI.yml")
+    mock(
+        GH.directory => Mock(([tagbot_yml, unrelated_yml], Dict())),
+        GH.file => Mock((; content=base64encode("uses: SciML/.github/.github/workflows/tagbot.yml@v1"))),
+    ) do directory, file
+        path, contents = TB.tagbot_file("SciML/CurveFit.jl")
```
DilumAluthge, line 60: Remind me: What does `TB.tagbot_file("SciML/CurveFit.jl")` do?

JamesWrigley: It will look through the repo files to check if there any actions that use tagbot, and if one is found then it will send the event to the repo. Previously it would look for the `JuliaRegistries/TagBot` string in files (or another `julia-ci` file), now it will also match for files named `tagbot.yml` (case-insensitive). That's needed for cases like CurveFit and the other SciML packages because they call a SciML org tagbot action rather than `JuliaRegistries/TagBot` directly.

DilumAluthge: Sorry, I'm asking about the tests specifically, not the source code changes. The source code changes look good to me - my only questions are about the tests.

JamesWrigley: Ah sorry :) The tests are mocking a repo containing two YAML files, one named `TagBot.yml` but without referencing `JuliaRegistries/TagBot`, and checking that `tagbot_file()` correctly finds the tagbot file so AutoMerge can send the event.

DilumAluthge: I guess what I'm asking is - does `TB.tagbot_file("SciML/CurveFit.jl")` fetch the live contents of the CurveFit.jl repo?

JamesWrigley: Ahh right, no, because the `GH.directory` and `GH.file` methods are mocked to return the test values.

JamesWrigley: But I can also change it to a dummy repo to make doubly sure?

DilumAluthge: No, I'm happy to use the mock. I was worried that we were hitting the live contents, which could cause our CI to break if the upstream repo is changed. The mock is better.

DilumAluthge (approved):

DilumAluthge:
CI is unhappy because of... something.

DilumAluthge:
Maybe not related to this PR.

JamesWrigley:
(bump)

JamesWrigley:
(bump)

ChrisRackauckas:
This wasn't released, can it get released?

JamesWrigley:
Yeah, current plan is to backport it to v10 since updating General to v11 would be a bit more work: #728.

PR #684, opened 2026-03-21
Update testitem-workflow URL

davidanthoff:
I moved the workflow into a different org, so this tracks the change.

PR #726, opened 2026-07-23
[AutoMerge] Bump version

JamesWrigley:
Hope it's ok :eyes: I'd like to get #724 in. I think v1.1 is the right bump since there were a few new features since the last release.


DilumAluthge:
So, even if we do register a new version of AutoMerge, it won't immediately affect General. We haven't upgraded General to v1 yet, and this would be blocked by that upgrade.

The upgrade of General to AutoMerge v1 is blocked by someone (me or Eric) being able to devote some time to it, but we've both been busy and haven't had time to dedicate to it.

JamesWrigley:
Ah I see, I didn't realize that. I'll see if I can look into it next week.

DilumAluthge:
As a short-term fix, perhaps we should backport the fixes you need to RegistryCI 10.x?

https://github.com/JuliaRegistries/RegistryCI.jl/pull/684#issuecomment-5138223453

PR #733, opened 2026-08-12
Support auto-merge for TZJData which uses build metadata 

omus:
Introduces a whitelist (`packages_with_build_metadata`) to allow specific packages which are using build metadata in their package registry version. The [TZJData.jl](https://github.com/JuliaTime/TZJData.jl) uses build metadata to communicate the IANA tzdata version (e.g. 2026c) without embedding that information in the major/minor/patch information.

Example of the auto-merge failure: https://github.com/JuliaRegistries/General/pull/164246#issuecomment-5265008673

PR #734, opened 2026-08-13
Allow `meets_sequential_version_number` to ignore build metadata

omus:
Split out from #733. Working around what is technically an issue with `VersionNumber` ordering in that build metadata should be ignored with determining precedence:

> Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85, 1.0.0+21AF26D3----117B344092BD.
> – https://semver.org/#spec-item-10

### DIFF
--- a/AutoMerge/src/guidelines.jl
+++ b/AutoMerge/src/guidelines.jl
@@ -813,7 +813,13 @@ function meets_sequential_version_number(
     else
         nextpatch(prv)
     end
-    ver <= nxt || return _invalid_sequential_version("version $ver skips over $nxt")
+
+    # Strip build metadata from `ver` as those fields should be ignored during this
+    # comparison. Without this versions which include build metadata will appear to
+    # skip over `nxt`.
+    clean_ver = VersionNumber(ver.major, ver.minor, ver.patch, ver.prerelease)
+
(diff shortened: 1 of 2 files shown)


AutoMerge/src/guidelines.jl
```diff
@@ -813,7 +813,13 @@ function meets_sequential_version_number(
     else
         nextpatch(prv)
     end
-    ver <= nxt || return _invalid_sequential_version("version $ver skips over $nxt")
+
+    # Strip build metadata from `ver` as those fields should be ignored during this
+    # comparison. Without this versions which include build metadata will appear to
+    # skip over `nxt`.
```
DilumAluthge, line 819:
```suggestion
    # Strip build metadata from `ver` as those fields should be ignored during this
    # comparison. Without this versions which include build metadata will appear to
    # skip over `nxt`.
    #
    # Note: We have a separate guideline for "non-JLL packages should not have
    # build metadata". That's a separate guideline from this one. So it's okay for
    # this guideline to strip metadata when doing the version comparison.
```

DilumAluthge:
One comment, and then after that this should be good to go.

DilumAluthge (approved):

